PR #11, opened 2017-09-18
[CHG] add a code to involvement category and a list of such involvement categories to membership requests

Olivier-LAURENT:


Olivier-LAURENT:
@EcoloDevelopper 

Thk in advance for your review

PR #52, opened 2018-03-21
[IMP] Simpler placeholders in email templates

andreparames:
refs #28441

PR #53, opened 2018-03-28
Involvements placeholders

andreparames:
refs #28616

(builds upon PR #52)

### DIFF
--- a/odoo_addons/mozaik_communication/email_coordinate.py
+++ b/odoo_addons/mozaik_communication/email_coordinate.py
@@ -22,7 +22,8 @@
 #     If not, see <http://www.gnu.org/licenses/>.
 #
 ##############################################################################
-from openerp import api, models
+from openerp import api, models, _
+from openerp.exceptions import Warning
 
 
 class EmailCoordinate(models.Model):
@@ -44,3 +45,15 @@ class EmailCoordinate(models.Model):
                 [('partner_id', 'in', partner_ids)])
         return mailing_contacts
 
+    @api.model
+    def get_url(self, path):
+        base_url = self.env['ir.config_parameter'].get_param(
+            'ecolo_website.base_url')
+        if not base_url:
+            raise Warning(
+                _('Please configure the base URL for the Ecolo website'))
+        if base_url.endswith('/'):
+            base_url = base_url[:-1]
+        if not path.startswith('/'):
+            path = '/' + path
+        return base_url + path
--- a/odoo_addons/mozaik_communication/email_coordinate.py
+++ b/odoo_addons/mozaik_communication/email_coordinate.py
@@ -30,21 +30,6 @@ class EmailCoordinate(models.Model):
 
     _inherit = 'email.coordinate'
 
-    @api.multi
-    @api.returns('mail.mass_mailing.contact')
-    def _get_linked_mailing(self):
-        '''
-        :rtype: [integer]
-        :rparam: list of `mail.mass_mailing.contact` associated with a partner
-            id into `ids`
-        '''
-        mailing_contacts = []
-        partner_ids = self.get_linked_partners()
-        if partner_ids:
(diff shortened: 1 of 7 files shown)


odoo_addons/mozaik_communication/email_coordinate.py
```diff
@@ -22,22 +22,38 @@
...
 
-class email_coordinate(orm.Model):
+class EmailCoordinate(models.Model):
 
-    def _get_linked_mailing(self, cr, uid, ids, context=None):
+    _inherit = 'email.coordinate'
+
+    @api.multi
+    @api.returns('mail.mass_mailing.contact')
+    def _get_linked_mailing(self):
```
Olivier-LAURENT: This method seems to be unused ! If right, it is better to remove it.

odoo_addons/mozaik_communication/models/email_template.py
```diff
@@ -0,0 +1,30 @@
...
+        'partner.involvement.category', string='Involvement Category',
+        domain=[('code', '!=', False)])
+
+    @api.onchange('placeholder_id')
+    def _onchange_placeholder_id(self):
+        super(EmailTemplate, self)._onchange_placeholder_id()
+        for wizard in self:
+            placeholder_value = wizard.placeholder_value
+            if placeholder_value and '{{CODE}}' in placeholder_value:
+                if not wizard.involvement_category_id:
```
Olivier-LAURENT: IMO, it would be better to make the onchange method also reactive on `involvement_category_id` and rebuild the `placeholder_value` anyway without raising error.

odoo_addons/mozaik_communication/models/email_template.py
```diff
@@ -0,0 +1,30 @@
...
+    from openerp.exceptions import Warning as UserError
+
+
+class EmailTemplate(models.Model):
+    _inherit = 'email.template'
+
+    # Fake field for auto-completing placeholder
+    involvement_category_id = fields.Many2one(
+        'partner.involvement.category', string='Involvement Category',
+        domain=[('code', '!=', False)])
```
Olivier-LAURENT, line 13: Can we add here `store=False` ? For a fake field it would be nice.

odoo_addons/mozaik_communication/wizard/distribution_list_mass_function_v8.py
```diff
@@ -14,13 +18,23 @@ class DistributionListMassFunction(models.TransientModel):
...
+        domain=[('code', '!=', False)])
 
     @api.onchange('placeholder_id')
     def _onchange_placeholder_id(self):
         for wizard in self:
             if wizard.placeholder_id:
-                wizard.placeholder_value = wizard.placeholder_id.placeholder
+                placeholder_value = wizard.placeholder_id.placeholder
                 wizard.placeholder_id = False
+                if '{{CODE}}' in placeholder_value:
```
Olivier-LAURENT: same as above

odoo_addons/mozaik_communication/models/email_template.py
Olivier-LAURENT: On the other hand it's not very compatible with your new module in the social repository. Thus forget my previous remark. Sorry.

Olivier-LAURENT (changes requested):

odoo_addons/mozaik_communication/email_coordinate.py
andreparames: I couldn't find any use either, so I've removed it.

odoo_addons/mozaik_communication/models/email_template.py
andreparames: I agree; thanks.

andreparames:
Thanks for the review, I've pushed an update.

Olivier-LAURENT (approved):
Thx for the job !

PR #54, opened 2018-03-28
[FIX] Show same wizard after saving as template

andreparames:


odoo_addons/mozaik_communication/wizard/distribution_list_mass_function_v8.py
```diff
@@ -29,7 +29,7 @@ def save_as_template(self):
             'view_mode': 'form',
             'view_type': 'form',
             'res_id': record.id,
```
Olivier-LAURENT: This `record` variable is in fact the iteration variable of the preceding loop.

Shouldn't be better to say first `self.ensure_one()` and use `self` everywhere in this method ?

Olivier-LAURENT (commented):
Hi @andreparames 
Just a non important comment.

odoo_addons/mozaik_communication/wizard/distribution_list_mass_function_v8.py
andreparames: I agree, thanks.

PR #56, opened 2018-04-10
[Proposal][CHG] Allow CSV queries to be extended

andreparames:
This is a proposal to make the CSV exporting more flexible, by creating a virtual ORM class that returns the CSV queries, instead of using constants, allowing them to be extended by other modules.


Olivier-LAURENT (approved):
Thx for the work.

PR #59, opened 2018-04-20
[CHG] Create A/B mechanism for distribution lists

andreparames:
refs #28891


odoo_addons/mozaik_communication/wizard/distribution_list_mass_function.py
```diff
@@ -284,6 +286,35 @@ def mass_function(self, cr, uid, ids, context=None):
                             cr, uid, [wizard.distribution_list_id.id],
                             context=context)
 
+                    if wizard.contact_ab_pc < 100:
```
Olivier-LAURENT: From the second batch, if user specify 100%, this condition is evaluated to False. It would be safer to say:
`if wizard.contact_ab_pc < 100 or context.get('mailing_group_id'):`

odoo_addons/mozaik_communication/mass_mailing.py
```diff
@@ -82,6 +82,8 @@ class MassMailing(orm.Model):
 
     _columns = {
```
Olivier-LAURENT: Entire class MassMailing here must be removed (its a duplicate of your new file  odoo_addons/mozaik_communication/models/mail_mass_mailing.py)

odoo_addons/mozaik_communication/models/mail_mass_mailing.py
```diff
@@ -0,0 +1,95 @@
...
+        Override this method to get resulting ids of the distribution list
+        """
+        if mailing.distribution_list_id:
+            if mailing.mailing_model == 'email.coordinate':
+                dl = mailing.distribution_list_id.with_context(
+                    main_object_field='email_coordinate_id',
+                    main_target_model='email.coordinate'
+                )
+                res = dl.get_complex_distribution_list_ids()[0]
+                if mailing.contact_ab_pc < 100 and mailing.group_id:
```
Olivier-LAURENT: As soon as a mass mailing is created with a group, this condition must be True. Thus it is better to say here:
`if mailing.contact_ab_pc < 100 or mailing.group_id:`
see also the related comment in the mass function wizard below

odoo_addons/mozaik_communication/models/mass_mailing_report.py
```diff
@@ -0,0 +1,37 @@
+# -*- coding: utf-8 -*-
+
+from openerp import models, fields, tools
+
+
+class MassMailingReport(models.Model):
+    _inherit = 'mail.statistics.report'
+    _auto = False
```
Olivier-LAURENT: Does this property need to be re-specified here?

odoo_addons/mozaik_communication/models/mail_mass_mailing.py
```diff
@@ -0,0 +1,95 @@
...
+from openerp import api, fields, models
+
+
+class MassMailing(models.Model):
+
+    _inherit = 'mail.mass_mailing'
+
+    create_uid =  fields.Many2one('res.users', readonly=True)
+    group_id = fields.Many2one(
+        'mail.mass_mailing.group', copy=True)
```
Olivier-LAURENT: a string property is needed here because it is possible to request an advanced "group by" from the searchbox

Olivier-LAURENT (changes requested):
hi @andreparames 
thx for the job... only a few comments

andreparames:
Thanks for the review! I've made the suggested changes.

Olivier-LAURENT (approved):

PR #71, opened 2018-05-25
[ADD] addon to send mail through the connector and to compute a job priority depending on the number of recipients of the mailing

Olivier-LAURENT:
refs #28990

### DIFF
--- a/odoo_addons/mail_job_priority/__init__.py
+++ b/odoo_addons/mail_job_priority/__init__.py
@@ -1,0 +1,2 @@
+from . import wizards
+from . import models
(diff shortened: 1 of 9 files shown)


sbidoul (commented):
This looks complex. It's still possible to change the oca module to handle priority from the context.

andreparames:
@sbidoul that would only make it possible to remove the three line method in `models/mail_mail.py`, no? Still, it could make sense to move it to the OCA addon.

sbidoul:
@andreparames the OCA module should definitely look for default_priority in the context when processing the create event.

sbidoul:
For the other method I'm not sure. Looking at the active_ids is a bit difficult to understand at first glance and looks like relying on an implementation detail of the mass mailing mechanism. I've the feeling it would be more readable if that was done at a higher level in the mass mailing module.

Also priority should maybe renamed to mail_job_priority or something to avoid ambiguity in the context variable name (priority is a bit generic for something in the context).

Issue #79, opened 2018-07-02
[MIG, V11] new module mozaik_abstract_model


Olivier-LAURENT:
Extract from mozaik_base features related to mozaik.abstract.model into a new module dedicated to it

Olivier-LAURENT:
Solved by #89

PR #89, opened 2018-07-31
[MIG][11.0] mozaik_abstract_model (from mozaik_base)

acsonefho:
- Migration of mozaik_abstract_model (from mozaik_base)
- Flake 8 (this module only)
- Pylint (this module only)
- Unit test added for this module

- New disable_tracking_installation module: extract behaviour from mozaik_abstract_model because (I think) the behaviour is different

**Depends on the branch of Denis**


odoo/addons/mozaik_abstract_model/models/mozaik_abstract_model.py
```diff
@@ -0,0 +1,177 @@
...
+        if mode == 'activate':
+            result.update({
+                'active': True,
+                'expire_date': False,
+            })
+        return result
+
+    @api.multi
+    @api.constrains('expire_date')
+    def _constrain_invalidate(self):
```
Olivier-LAURENT: Maybe better to name constrains methods with `_check` prefix

odoo/addons/mozaik_abstract_model/models/mozaik_abstract_model.py
```diff
@@ -0,0 +1,177 @@
...
+        Reactivates an object by setting
+        :param vals: dict
+        :return: bool
+        """
+        vals = vals or {}
+        vals.update(self.get_fields_to_update('activate'))
+        return self.write(vals)
+
+    @api.model
+    def get_fields_to_update(self, mode):
```
Olivier-LAURENT: It would be better to underscore this method (i.e. not useful directly from JS). Correct ?

odoo/addons/mozaik_abstract_model/models/mozaik_abstract_model.py
```diff
@@ -0,0 +1,177 @@
...
+                    })
+                if self._inactive_cascade:
+                    self._invalidate_active_relations()
+        return super(MozaikAbstractModel, self).write(vals)
+
+    @api.multi
+    def action_invalidate(self, vals=None):
+        """
+        Invalidates an object
+        Note: Argument vals must be the last in the signature
```
Olivier-LAURENT: Obsolete note probably

odoo/addons/mozaik_abstract_model/models/mozaik_abstract_model.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+class MozaikAbstractModel(models.AbstractModel):
+    _name = 'mozaik.abstract.model'
+    _inherit = ['mail.thread']
+    _description = 'Mozaik abstract model'
+
+    _allowed_inactive_link_models = []
+    _inactive_cascade = False
+    _unicity_keys = 'N/A'
```
Olivier-LAURENT: It would be better to keep `None` here as default value

odoo/addons/mozaik_abstract_model/models/mozaik_abstract_model.py
```diff
@@ -0,0 +1,177 @@
...
+        default=True,
+    )
+
+    def init(self):
+        """
+        Create unit index based on models who implements this abstract model.
+        :return:
+        """
+        result = super(MozaikAbstractModel, self).init()
+        if not self._auto or self._unicity_keys == 'N/A':
```
Olivier-LAURENT: Definitively if model is not `_auto`, nothing of this method can be executed !

Olivier-LAURENT (changes requested):

odoo/addons/mozaik_abstract_model/models/mozaik_abstract_model.py
acsonefho: Actually it's the case: it returns directly the result of the super()

Olivier-LAURENT:
close #79

PR #90, opened 2018-08-02
[MIG][11.0] mozaik_base

acsonefho:
- Migration of mozaik_base
- Flake 8 (this module only)
- Pylint (this module only)

**Update mozaik_abstract_model:**
- Implements new unit test about validation/invalidate

**Depends on:**
- Branch of Denis
- mozaik_abstract_model


odoo/addons/mozaik_abstract_model/models/mozaik_abstract_model.py
```diff
@@ -153,7 +160,8 @@ def _check_invalidate(self):
                     _logger.info(
                         'Remaining active m2o for %s(%s): %s',
                         self._name, k, v)
-                raise ValidationError(INVALIDATE_ERROR)
+                raise ValidationError(_('Impossible invalidation: at least '
```
Olivier-LAURENT: maybe  better to say : `Invalidation not allowed: ...`

odoo/addons/mozaik_abstract_model/tests/common_mozaik_abstract_model.py
```diff
@@ -82,3 +87,42 @@ def test_disable_cascade2(self):
...
+        :return: bool
+        """
+        # Ensure test correctly implemented
+        self.assertIsInstance(self.validates, models.BaseModel)
+        result = self.invalidate_success.action_invalidate()
+        self.assertTrue(result)
+        for inv_success in self.invalidate_success:
+            self.assertFalse(inv_success.active)
+            self.assertTrue(bool(inv_success.expire_date))
+        return True
```
Olivier-LAURENT: With `True` or not finally the `return` ?

Olivier-LAURENT (changes requested):
In addition, in `abstract_model` it would be nice to `copy=False` for `active` and `expire_date`

PR #105, opened 2018-08-27
[MIG] mozaik_person

Olivier-LAURENT:


odoo/addons/mozaik_duplicate/models/abstract_duplicate.py
```diff
@@ -73,11 +73,14 @@ def write(self, vals):
...
         """
         trigger_fields = self._get_trigger_fields(vals.keys())
-        result = super().write(vals)
-        if self and trigger_fields and \
-                not self._context.get('escape_detection'):
+        detect = self and trigger_fields and \
+            not self._context.get('escape_detection')
+        if detect:
             self_suspend = self.suspend_security()
             values = [d._get_discriminant_value() for d in self_suspend]
```
acsonefho, line 80: This line can be deleted? Because it's related to the same line after the super (into the if)?

odoo/addons/mozaik_person/models/res_partner.py
```diff
@@ -1,32 +1,245 @@
...
+
+    @api.multi
+    @api.constrains('identifier')
+    def _check_identifier(self):
+        """
+        Check if identifier is unique
+        """
+        other = self.env['res.partner'].sudo()
+        for partner in self.filtered(lambda s: s.identifier):
+            p = other.search([
```
acsonefho: I think it should be better (about perf) to put the search before the loop.
A domain like this:
[('identifier', 'in', self.mapped("identifier")]
(you can't ignore current ids otherwise you'll ignore too much partners.
And into the loop, you can use a filtered like this:
(result is the result of the search using previous domain)
if result.filtered(lambda p: p != partner and p.identifier == partner.identifier):
    raise exception... (duplicates found)

odoo/addons/mozaik_person/models/res_partner.py
```diff
@@ -1,32 +1,245 @@
...
+            'partner_id': self.id,
+            'login': login,
+        })
+
+        user = self.env['res.users'].with_context(
+            no_reset_password=True).create(vals)
+
+        return user
+
+    def _update_identifier_sequence(self):
```
acsonefho: I think this function is not used anymore (only in unit test)

acsonefho (changes requested):
Minor changes
Enjoy :)

odoo/addons/mozaik_duplicate/models/abstract_duplicate.py
Olivier-LAURENT: Hummm ! But the `write()` has maybe touched the discriminant value !

Olivier-LAURENT: @acsonefho 
Hummm ! But the write() that follows will maybe touch the discriminant value !

odoo/addons/mozaik_person/models/res_partner.py
Olivier-LAURENT: @acsonefho 
This is a post customer data  recovery method (to manually call) to re-initialize the next number of the sequence. I'm going to add a 'one-shot' cron.

PR #116, opened 2018-09-07
[MIG] mozaik_membership

RobinetDenisAcsone:


odoo/addons/mozaik_membership/models/postal_coordinate.py
```diff
@@ -1,107 +1,59 @@
...
-                      ('id', 'in', ids)],
-            context=context)
-        res = super(postal_coordinate, self).write(
-            cr, uid, ids, vals, not_main_ids=not_main_ids, context=context)
+        res = super().write(vals)
         if vals.get('is_main', False):
-            self._update_partner_int_instance(cr, uid, ids, context=context)
-            self.update_notify_followers(
-                cr, uid, vals, not_main_ids, ids=ids, context=context)
+            self._update_partner_int_instance()
```
acsonefho: As this IF is the same for create/write, maybe we can group them into 1 function?

odoo/addons/mozaik_membership/models/res_partner.py
```diff
@@ -88,98 +76,25 @@ def _compute_current_membership_line_id(self):
                 lambda s: s.active) or False
```
acsonefho: For security during the compute, maybe it should be better to take only the first membership_line_ids (to ensure 1 result max).
Because if more than 1 is active = True, it could have an error.
You can import odoo.fields import first
Then: partner.current_membership_line_id = first(current_membership_line_id)
(but we have to remove also the or False on the filtered)

odoo/addons/mozaik_membership/models/res_partner.py
```diff
@@ -218,185 +133,103 @@ def _update_membership_line(self, ref=False, date_from=False):
...
+        self.ensure_one()
+        user = super()._create_user(login, group_ids)
+        self.int_instance_m2m_ids = self.int_instance_id
+        return user
+
+    @api.multi
+    @api.depends("membership_line_ids", "membership_line_ids.product_id")
+    def _compute_subscription_product_id(self):
+        for partner in self:
+            membership_line = partner.membership_line_ids.filtered("active")
```
acsonefho: Same remark here, maybe use the first(...) to ensure 1 result max (for membership_line)

odoo/addons/mozaik_membership/models/res_users.py
```diff
@@ -1,69 +1,43 @@
...
-        """
-        ass_ids = self.internal_assemblies(cr, uid)
-        if not ass_ids:
-            return []
-        dom = [(
-            'int_assembly_id', 'in', ass_ids
-        )]
-        ids = self.pool['int.mandate'].search(cr, SUPERUSER_ID, dom)
-        return ids
+        instances = self.env['int.instance'].sudo().search(dom)
```
acsonefho: I think it could be replaced by (as a special sudo env already exists):
self_sudo.env['int.instance'].search(dom)

odoo/addons/mozaik_membership/models/structure.py
```diff
@@ -1,185 +1,164 @@
...
-from openerp.osv import orm
+from odoo import api, models, fields, SUPERUSER_ID
+from odoo.fields import first
+from odoo.tools.safe_eval import safe_eval
 
-PARTNER_ACTION = 'mozaik_person.natural_res_partner_action'
 
-
-class sta_assembly(orm.Model):
+class StaAssembly(models.Model):
```
acsonefho: The class name doesn't match with the python file name

odoo/addons/mozaik_membership/models/structure.py
```diff
@@ -1,185 +1,164 @@
...
-        vals.update(self._pre_update(cr, uid, vals, context=context))
-        res = super(sta_assembly, self).write(
-            cr, uid, ids, vals, context=context)
+        vals.update(self._pre_update(vals))
+        res = super().write(vals)
         return res
 
 
-class int_instance(orm.Model):
+class IntInstance(models.Model):
```
acsonefho: This class should be extracted from the file I think (1 python file per model)

odoo/addons/mozaik_membership/models/structure.py
```diff
@@ -1,185 +1,164 @@
...
     @api.multi
     def _get_member_ids(self):
+        self.ensure_one()
         partner_obj = self.env['res.partner']
         domain = [
             ('int_instance_id', '=', self.id),
             ('is_company', '=', False)
         ]
-        return [partner.id for partner in partner_obj.search(domain)]
+        return partner_obj.search(domain)
```
acsonefho: Instead of search (who force a SQL query), it should be better to use the O2M fields (partner_ids) and then a filter (with is_company field). So if the partner is already in the cache, we don't trigger a new SQL query.
And this function is used into a compute function into a loop. So we'll have 1 sql query per record into the self. So I think the filtered is the best solution about perf

odoo/addons/mozaik_membership/models/structure.py
```diff
@@ -1,185 +1,164 @@
...
         Returns the default Internal Instance
         """
-        res = [super(int_instance, self).get_default(cr, uid, context=context)]
-        if res:
-            res = self.search(cr, uid, [('id', 'in', res)], context=context)
+        res = super()._get_default_int_instance()
         if not res:
-            res = self.pool['res.users'].internal_instances(cr, uid)
-        return res and res[0] or False
+            res = self.pool['res.users']._internal_instances()
```
acsonefho: pool should be replaced by 'env'

odoo/addons/mozaik_membership/models/structure.py
```diff
@@ -1,185 +1,164 @@
...
+        return first(res)
 
 
-class int_assembly(orm.Model):
+class IntAssembly(models.Model):
```
acsonefho: This class should be into another python file

odoo/addons/mozaik_membership/models/structure.py
```diff
@@ -1,185 +1,164 @@
...
 
-    def create(self, cr, uid, vals, context=None):
+    @api.model
+    def create(self, vals):
         '''
         Responsible Internal Instance linked to the result Partner is the
         Instance of the Assembly
         '''
         if 'instance_id' in vals:
             vals.update({'int_instance_id': vals['instance_id']})
```
acsonefho: As this if is the same into create and write, maybe it could be extracted into a same function

odoo/addons/mozaik_membership/models/structure.py
```diff
@@ -1,185 +1,164 @@
...
-
-    def unlink(self, cr, uid, ids, context=None):
-        res = super(int_assembly, self).unlink(cr, uid, ids, context=context)
-        self.pool['ir.rule'].clear_cache(cr, uid)
+        res = super().write(vals)
         return res
 
 
-class ext_assembly(orm.Model):
+class ExtAssembly(models.Model):
```
acsonefho: Same here: class should be extracted into a new python file

odoo/addons/mozaik_membership/models/structure.py
```diff
@@ -1,185 +1,164 @@
...
     _inherit = 'ext.assembly'
 
-    def create(self, cr, uid, vals, context=None):
+    @api.model
+    def create(self, vals):
         '''
         Responsible Internal Instance linked to the result Partner is
         the Instance of the Assembly
         '''
         if 'instance_id' in vals:
```
acsonefho: This if is quite similar for few models. Maybe we can create an abstract model who is doing this if for every create/write.
@Olivier-LAURENT what do you think? Bad/good idea, without increase the complexity and understanding?

odoo/addons/mozaik_membership/models/virtual_models.py
```diff
@@ -1,11 +1,11 @@
...
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from openerp.osv import orm, fields
-from openerp import tools
+from odoo import models, fields
+from odoo import tools
 
 
-class virtual_master_partner(orm.Model):
+class VirtualMasterPartner(models.Model):
```
acsonefho: File name doesn't match with the model/class name

odoo/addons/mozaik_membership/models/virtual_models.py
```diff
@@ -118,18 +112,18 @@ def init(self, cr):
             )""")
 
 
-class virtual_custom_partner(orm.Model):
+class VirtualCustomPartner(models.Model):
```
acsonefho: This model should be extracted (1 file per model)

odoo/addons/mozaik_membership/tests/test_force_int_instance.py
```diff
@@ -1,72 +1,36 @@
...
-#     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#     GNU Affero General Public License for more details.
-#
-#     You should have received a copy of the
-#     GNU Affero General Public License
-#     along with mozaik_membership.
-#     If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
+# Copyright 2017 ACSONE SA/NV
```
acsonefho: Replace 2017 per 2018

odoo/addons/mozaik_membership/tests/test_postal_coordinate.py
```diff
@@ -63,63 +28,57 @@ def test_change_co_residency_rights(self):
...
 
-        cor_ids = cr_mod.search(cr, uid, [('address_id', '=', new_adr_id)])
+        cor_ids = cr_mod.search([('address_id', '=', new_adr.id)])
         self.assertFalse(cor_ids)
-        pc_ids = pc_mod.search(cr, uid, [('address_id', '=', new_adr_id)])
+        pc_ids = pc_mod.search([('address_id', '=', new_adr.id)])
         self.assertFalse(pc_ids)
 
-        postal_coordinates_ids = []
+        postal_coordinates = self.env["postal.coordinate"]
```
acsonefho: Don't forget the ".browse()" on this line to avoid mix between "models" and recordset

odoo/addons/mozaik_membership/tests/test_postal_coordinate.py
```diff
@@ -63,63 +28,57 @@ def test_change_co_residency_rights(self):
...
-        partners = ['%s.res_partner_marc',
-                    '%s.res_partner_jacques',
-                    '%s.res_partner_paul']
-        partner_ids = []
-        for partner in partners:
-            partner_ids.append(self.ref(partner % self._module_ns))
+        partners_xmlid = ['mozaik_coordinate.res_partner_marc',
+                          'mozaik_coordinate.res_partner_jacques',
+                          'mozaik_coordinate.res_partner_paul']
+        partners = self.env["res.partner"]
```
acsonefho: Same remark here, add ".browse()"

odoo/addons/mozaik_membership/tests/test_res_partner.py
```diff
@@ -78,371 +41,37 @@ def get_partner(self, partner_id=False):
         # check each time the current state change
         return self.partner_obj.browse(self.cr, self.uid, partner_id)
```
acsonefho: Should be replaced with new API (or maybe delete this function if not used anymore)

odoo/addons/mozaik_membership/tests/test_structure.py
```diff
@@ -1,137 +1,99 @@
...
-        assembly = sta_assembly_model.browse(cr, uid, sta_id, context=context)
-        self.assertEqual(assembly.partner_id.int_instance_id.id,
-                         assembly.instance_id.int_instance_id.id,
-                         'Create state assembly fails with wrong internal \
-                         instance linked to the result partner')
+        self.assertTrue(assembly.partner_id.is_assembly)
 
         # 2.2/ Update the assembly
-        instance_id = self.ref('%s.sta_instance_09' % self._module_ns)
+        # instance_id = self.env.ref('mozaik_structure.sta_instance_03')
```
acsonefho: Maybe could be removed

odoo/addons/mozaik_membership/views/res_partner.xml
```diff
@@ -112,93 +112,86 @@
...
-                <attribute name="groups">mozaik_base.mozaik_res_groups_officer</attribute>
-            </xpath>
-
-            <xpath expr="//header/div/button[@name='action_invalidate']" position="after">
-                <button name="action_invalidate" string="Invalidate"
-                        type="object" confirm="Invalidate contact: proceed anyway?"
-                        attrs="{'invisible': [('kind', '!=', 'p')]}"
-                        invisible="context.get('in_mozaik_officer') or not context.get('in_mozaik_user')"/>
-            </xpath>
+            <!--<xpath expr="//header/div/button[@name='action_invalidate']" position="after">-->
```
acsonefho: Maybe could be removed?

odoo/addons/mozaik_membership/wizards/change_main_address.py
```diff
@@ -1,96 +1,88 @@
...
-    def onchange_address_id(self, cr, uid, ids, address_id,
-                            old_int_instance_id, context=None):
-        res = {}
+    @api.onchange("address_id")
+    def _onchange_address_id(self):
+        self.ensure_one()
         new_int_instance_id = False
         keeping_mode = 3
-        if not old_int_instance_id:
+        if not self.old_int_instance_id:
```
acsonefho, line 58: I think this old_int_instance_id field is only used for the onchange, to know if the int_instance_id has been updated from the original value (so the value into the DB).
If it's the case, we can remove this field, because into the onchange, we can have access to the real value (into DB) and check if the value is different that the "onchange value".
@Olivier-LAURENT Maybe you can answer to this question, about the real utility of this "old" field.

odoo/addons/mozaik_membership/wizards/generate_reference.py
```diff
@@ -16,85 +15,84 @@
...
+            partner_ids = self.env.context.get('active_ids')
 
         if partner_ids:
             # search member with reference
             member_ids = partner_obj.search(
-                cr, uid, [('membership_state_id.code', '=', 'member'),
-                          ('id', 'in', partner_ids),
-                          ],
-                context=context)
+                [('membership_state_id.code', '=', 'member'),
```
acsonefho: Maybe we can use a O2M field with a filtered? To avoid SQL query?

odoo/addons/mozaik_membership/wizards/generate_reference.py
```diff
@@ -16,85 +15,84 @@
...
+                 ('id', 'in', partner_ids)])
             candidate_ids = partner_obj.search(
-                cr, uid, [('membership_state_id.code', '=',
-                           'member_candidate'),
-                          ('id', 'in', partner_ids),
-                          ],
-                context=context)
+                [('membership_state_id.code', '=', 'member_candidate'),
```
acsonefho: Same remark here

odoo/addons/mozaik_membership/wizards/generate_reference.py
```diff
@@ -16,85 +15,84 @@
...
+                 ('id', 'in', partner_ids)])
             former_ids = partner_obj.search(
-                cr, uid, [('membership_state_id.code', '=',
-                           'former_member'),
-                          ('id', 'in', partner_ids),
-                          ],
-                context=context)
+                [('membership_state_id.code', '=', 'former_member'),
+                 ('id', 'in', partner_ids)])
```
acsonefho: Same remark here again

odoo/addons/mozaik_membership/wizards/generate_reference.py
```diff
@@ -104,41 +102,33 @@ def generate_reference(self, cr, uid, ids, context=None):
...
-        for wiz in self.browse(cr, uid, ids, context=context):
-            partner_ids = eval(wiz.partner_ids)
-            session = ConnectorSession(cr, uid, context=context)
+        for wiz in self:
+            partner_ids = wiz.partner_ids
             if len(partner_ids) > worker_pivot:
-                generate_reference_action.delay(
-                    session, self._name, partner_ids, wiz.reference_date,
-                    context=context)
+                self.with_delay().generate_reference_action(
```
acsonefho: Could be useful to add a  little description (with description parameter) into the with_delay() to know exactly what does the new job

odoo/addons/mozaik_membership/wizards/pass_former_member.py
```diff
@@ -3,40 +3,43 @@
...
-            partner_ids = context.get('active_ids')
+        if self.env.context.get('active_domain'):
+            active_domain = self.env.context.get('active_domain')
+            partner_ids = partner_obj.search(active_domain)
+        elif self.env.context.get('active_ids'):
+            partner_ids = self.env.context.get('active_ids')
 
         if partner_ids:
             # search member with reference
             domain = [('id', 'in', partner_ids), ('reference', '!=', False)]
```
acsonefho: We can have an issue here.
Because few lines before, partner_ids could be a recordset or a list of int.
If partner_ids is a recordset, the search will raise an exception because it need list of int.
I think the line 32 (partner_ids = partner_obj.search(active_domain)) should be replaced by a filtered. And do the search into the elif with the active_ids.

odoo/addons/mozaik_membership/wizards/pass_former_member.py
```diff
@@ -45,33 +48,30 @@ def _get_selected_values(self, cr, uid, context=None):
...
+                          'concerned_partner_ids',
+                          'go']
+        if any(el in fields_list for el in current_fields):
+            partner_ids, concerned_partner_ids, concerned_members = \
+                self._get_selected_values()
+            res['nb_selected'] = len(partner_ids)
+            res['concerned_members'] = concerned_members
+            res['concerned_partner_ids'] = [(6, 0, concerned_partner_ids.ids)]
+            curr_month = date.today().month
+            date_ok = curr_month in AVAILABLE_MONTHS
```
acsonefho: Maybe we can put AVAILABLE_MONTHS into a function (with api.model) to avoid global variables.

odoo/addons/mozaik_membership/wizards/pass_former_member.py
```diff
@@ -81,42 +81,26 @@ def pass_former_member(self, cr, uid, ids, context=None):
...
-    pass
+    @job(default_channel="root.pass_former_member")
+    def pass_former_member_action(self, partner_ids):
+        """
+        Pass to former Member for each partner
+        Reset reference for each other
+        """
+        partner_ids.decline_payment()
+        domain = [('id', 'in', partner_ids.ids), ('reference', '!=', False)]
+        partners = self.env["res.partner"].search(domain)
```
acsonefho: Maybe better to use filtered here too to avoid SQL query

acsonefho (changes requested):
Enjoy :)

odoo/addons/mozaik_membership/wizards/generate_reference.py
RobinetDenisAcsone: O2M isn't possible (api.model) but a filtered is now used.

odoo/addons/mozaik_membership/wizards/change_main_address.py
RobinetDenisAcsone: ```old_int_instance_id``` is the instance of the partner, if we remove it from the wizard, we need to add a relation to the partner or retrieved this partner from the context

odoo/addons/mozaik_membership/models/int_instance.py
```diff
@@ -0,0 +1,66 @@
...
+    partner_m2m_ids = fields.Many2many(
+        comodel_name="res.partner")
+
+    @api.multi
+    @api.depends("partner_ids")
+    def _compute_member_count(self):
+        for inst in self:
+            inst.member_count = len(self.partner_ids.filtered(
+                lambda p, s:
+                p.int_instance_id == s.id and not p.is_company))
```
Olivier-LAURENT: Is this lambda really valid ?  ... you compare recordset against id .... and `s` is not defined ... I propose:
`lambda s, ii=inst: s.int_instance_id == ii and not s.is_company`

odoo/addons/mozaik_membership/models/abstract_coordinate.py
```diff
@@ -0,0 +1,34 @@
...
+            fol = coord.partner_instance_id._get_instance_followers()
+            coord.message_subscribe(fol.ids)
+
+    @api.model
+    def create(self, vals):
+        '''
+        update and notify follower if is_main is True
+        '''
+        res = super().create(vals)
+        res.sudo()._update_followers()
```
Olivier-LAURENT: Shouldn't this method also called from `write()` ?  What about the followers after a change of the int_instance_id of the partner ?

odoo/addons/mozaik_membership/models/int_instance.py
```diff
@@ -0,0 +1,66 @@
...
+    def get_member_action(self):
+        self.ensure_one()
+        action = self.env.ref(
+            "mozaik_person.res_partner_natural_person_action").read()[0]
+        domain = safe_eval(action['domain'])
+        action['domain'] = domain + [('int_instance_id', '=', self.id)]
+        return action
+
+    @api.model
+    def check_mail_message_access(self, res_ids, operation, model_name=None):
```
Olivier-LAURENT: In which case this method is useful ? Without answer I suggest to remove it. If any, later, we will find the appropriate solution.

odoo/addons/mozaik_membership/models/int_instance.py
```diff
@@ -0,0 +1,66 @@
...
+        sufficient rights to create the related notification
+        """
+        if operation == 'create':
+            return
+        super().check_mail_message_access(res_ids, operation, model_name)
+
+    @api.model
+    def create(self, vals):
+        if not vals.get('parent_id'):
+            if self.env.uid != SUPERUSER_ID:
```
Olivier-LAURENT: `if not self.env.user.is_admin()`

odoo/addons/mozaik_membership/models/int_instance.py
```diff
@@ -0,0 +1,66 @@
...
+        return res
+
+    @api.model
+    def _get_default_int_instance(self):
+        """
+        Returns the default Internal Instance
+        """
+        res = super()._get_default_int_instance()
+        if not res:
+            res = self.env['res.users']._internal_instances()
```
Olivier-LAURENT: it would be better to make `first()` here...

odoo/addons/mozaik_membership/models/int_instance.py
```diff
@@ -0,0 +1,66 @@
...
+        return first(res)
```
Olivier-LAURENT: ... and not here !

odoo/addons/mozaik_membership/models/postal_coordinate.py
```diff
@@ -1,107 +1,62 @@
...
             if pc.is_main and pc.active == pc.partner_id.active and \
                     pc.partner_id.membership_state_id:
                 partner = pc.partner_id
                 cur_int_instance_id = partner.int_instance_id.id
-                def_int_instance_id = self.pool['int.instance'].\
-                    get_default(cr, uid)
+                def_int_instance_id = self.env['int.instance']\
+                    ._get_default_int_instance()
                 # get instance_id of address or keep default otherwise
                 zip_id = pc.address_id.address_local_zip_id
```
Olivier-LAURENT: `city_id` ??

odoo/addons/mozaik_membership/models/postal_coordinate.py
```diff
@@ -1,107 +1,62 @@
...
-            'partner_id', 'kind', string='Partner Kind',
-            type='selection', selection=AVAILABLE_PARTNER_KINDS,
-            store=_partner_kind_store_trigger),
-    }
-
-    def _update_partner_int_instance(self, cr, uid, ids, context=None):
+    @api.multi
+    def _update_partner_int_instance(self):
         """
         Update instance of partner linked to the postal coordinate case where
```
Olivier-LAURENT: Rewording...
Update instance of partner linked to the postal coordinate
when it becomes main and its `active` field is equal to the `active` field of its partner.
Instance is the default one if no instance found on the city

odoo/addons/mozaik_membership/models/postal_coordinate.py
```diff
@@ -1,107 +1,62 @@
...
                 partner = pc.partner_id
                 cur_int_instance_id = partner.int_instance_id.id
-                def_int_instance_id = self.pool['int.instance'].\
-                    get_default(cr, uid)
+                def_int_instance_id = self.env['int.instance']\
+                    ._get_default_int_instance()
                 # get instance_id of address or keep default otherwise
                 zip_id = pc.address_id.address_local_zip_id
                 new_int_instance_id = \
                     zip_id and zip_id.int_instance_id.id or \
```
Olivier-LAURENT: remove `.id`

odoo/addons/mozaik_membership/models/postal_coordinate.py
```diff
@@ -1,107 +1,62 @@
...
-                        cr, uid, partner.id, new_int_instance_id,
-                        context=context)
+                    partner._change_instance(new_int_instance_id)
 
-    def create(self, cr, uid, vals, context=None):
-        '''
+    @api.model
+    def create(self, vals):
+        """
         call `_update_partner_int_instance` if `is_main` is True
```
Olivier-LAURENT: Compute followers and change partner instance if any

odoo/addons/mozaik_membership/models/postal_coordinate.py
```diff
@@ -1,107 +1,62 @@
...
-        '''
-        context = context or {}
-        change_instance = not context.get('keep_current_instance')
-        ctx = dict(context, delay_notification=change_instance)
-        res = super(postal_coordinate, self).create(
-            cr, uid, vals, context=ctx)
+        """
+        change_instance = not self.env.context.get('keep_current_instance')
+        self_ctx = self.with_context(delay_notification=change_instance)
```
Olivier-LAURENT: This context is unused in this addon. I suggest to remove it here.

odoo/addons/mozaik_membership/models/postal_coordinate.py
```diff
@@ -1,107 +1,62 @@
...
-            self.update_notify_followers(
-                cr, uid, vals, [res], context=context)
+            res._update_postal_follower()
         return res
 
-    def write(self, cr, uid, ids, vals, context=None):
+    @api.model
+    def write(self, vals):
         '''
         call `_update_partner_int_instance` if `is_main` is True
```
Olivier-LAURENT: Recompute followers and change partner instance if any

odoo/addons/mozaik_membership/models/res_partner.py
```diff
@@ -80,106 +65,37 @@ class ResPartner(models.Model):
...
         Compute the kind of partner, computed field used in ir.rule
         '''
-        flds = [
-            'is_assembly', 'is_company', 'identifier', 'membership_state_code',
-        ]
-        res = {}
-        vals = self.read(cr, SUPERUSER_ID, ids, flds, context=context)
-        for val in vals:
-            if val['is_assembly']:
+        for val in self:
```
Olivier-LAURENT: `val` is no longer an appropriate name now

odoo/addons/mozaik_membership/models/res_partner.py
```diff
@@ -218,185 +134,104 @@ def _update_membership_line(self, ref=False, date_from=False):
...
         fol_ids = []
-        for pid in ids:
-            fol_ids = self._get_followers_assemblies(
-                cr, uid, pid, context=ctx)
-            self.message_subscribe(
-                cr, uid, [pid], fol_ids, context=context)
+        for partner in self:
+            fol_ids = self.with_context(new_instance_id=int_instance_id)\
+                ._get_followers_assemblies()
+            partner.message_subscribe(fol_ids.ids)
```
Olivier-LAURENT: `partner_ids=...`

odoo/addons/mozaik_membership/models/res_partner.py
```diff
@@ -218,185 +134,104 @@ def _update_membership_line(self, ref=False, date_from=False):
...
 
-    def create(self, cr, uid, vals, context=None):
+    @api.model
+    def create(self, vals):
         '''
         If partner has an identifier then update its followers
         '''
         if not vals.get('is_company', False):
             # Force the state here to avoid a security alert
             # when creating the workflow and updating the first time
```
Olivier-LAURENT: this comment seems out-of-date (regarding workflow)
I suggest report this business rule in the default property of the field `membership_state_id`

odoo/addons/mozaik_membership/models/res_partner.py
```diff
@@ -218,185 +134,104 @@ def _update_membership_line(self, ref=False, date_from=False):
...
+            state_obj = self.env['membership.state']
             vals.update({
-                'membership_state_id': state_obj._state_default_get(
-                    cr, uid, context=context),
+                'membership_state_id': state_obj._get_default_state().id,
             })
-        res = super(res_partner, self).create(cr, uid, vals, context=context)
-        if vals.get('identifier', 0) > 0:
+        res = super().create(vals)
+        if 'identifier' in vals and res.identifier > 0:
```
Olivier-LAURENT: remove the `and res.identifier > 0` part of the expression together with the following comment (it was a legacy rule)

odoo/addons/mozaik_membership/models/res_partner.py
```diff
@@ -218,185 +134,104 @@ def _update_membership_line(self, ref=False, date_from=False):
...
             return False
-        fol_ids = ia_obj.get_followers_assemblies(
-            cr, uid, int_instance_id, context=context)
+        fol_ids = int_instance._get_instance_followers()
         return fol_ids or False
 
-    def _subscribe_assemblies(
-            self, cr, uid, ids, int_instance_id=None, context=None):
+    @api.multi
+    def _subscribe_assemblies(self, int_instance_id=None):
```
Olivier-LAURENT, line 156: Is this method still called ? However, it is this method that subscribes followers on partners... ?

odoo/addons/mozaik_membership/models/res_partner.py
```diff
@@ -445,63 +276,182 @@ def write(self, vals):
                 date_from=data.get('accepted_date'))
 
         # update partners followers
-        p_upd_fol.sudo()._update_followers()
+        if vals.get('int_instance_id'):
+            p_upd_fol.sudo()._update_followers()
```
Olivier-LAURENT: It appears all these kind of calls use sudo(), here and in coordinates. The real leagcy reason is the reading of instances "above" the level of the user (in the instances pyramid). Thus, (maybe), it would be better to improve this by sudoing the reading method (located in mozaik_strcture/int_instance.py), and leaving all `_update_followers()` call without invoking `sudo()`. What do you think ?

odoo/addons/mozaik_membership/models/res_partner.py
```diff
@@ -445,63 +276,182 @@ def write(self, vals):
...
+            self.update_state("member_committee")
+
+    @api.multi
+    def update_state(self, membership_state_code):
+        """
+        :type membership_state_code: char
+        :param membership_state_code: code of `membership.state`
+        """
+        self.ensure_one()
+        if not self.env.context.get('lang'):
```
Olivier-LAURENT: legacy check ... because workflow didn't propagate context... IMO no more useful

odoo/addons/mozaik_membership/models/res_partner.py
```diff
@@ -445,63 +276,182 @@ def write(self, vals):
...
+        return res
+
+    def paid(self):
+        if self.membership_state_code == "former_member":
+            self.update_state("former_member_committee")
+        elif self.membership_state_code != "member":
+            self.update_state("member_committee")
+
+    @api.multi
+    def update_state(self, membership_state_code):
```
Olivier-LAURENT: _update_state()

odoo/addons/mozaik_membership/models/res_partner.py
```diff
@@ -445,63 +276,182 @@ def write(self, vals):
...
+            ],
+            'reference': False,
+            'amount': False,
+        }
+
+        if membership_state_code == 'supporter':
+            vals['free_member'] = True
+
+        if membership_state_code == 'member_candidate':
+            vals['del_doc_date'] = False
```
Olivier-LAURENT: Is this field still exist ?

odoo/addons/mozaik_membership/models/res_users.py
```diff
@@ -1,69 +1,43 @@
...
 
-class res_users(orm.Model):
+class ResUsers(models.Model):
 
     _inherit = 'res.users'
 
     int_instance_m2m_ids = fields.Many2many(
-        related='partner_id.int_instance_m2m_ids', inherited=True)
+        related='partner_id.int_instance_m2m_ids')
+    lang = fields.Selection(related='partner_id.lang')
```
Olivier-LAURENT: Who does really use this field ?

Olivier-LAURENT (changes requested):

odoo/addons/mozaik_membership/models/res_partner.py
RobinetDenisAcsone: I'm agreed, changed

odoo/addons/mozaik_membership/models/int_instance.py
RobinetDenisAcsone: I didn't find it useful, removed

PR #120, opened 2018-09-19
[IMP] Improve the error message

acsonefho:
Improve error message during `create`/`write` of models who inherit `mozaik.abstract.model`
So, models have to overwrite the `_get_exception_messages()` function to specify a message for a specific SQL constraint name.
Also, add the function `_get_index_name()` to have the generated constraint name.

### DIFF
--- a/odoo/addons/mozaik_abstract_model/models/mozaik_abstract_model.py
+++ b/odoo/addons/mozaik_abstract_model/models/mozaik_abstract_model.py
@@ -2,7 +2,8 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 import logging
 from psycopg2.extensions import AsIs
-from odoo import api, fields, models, _
+from psycopg2 import IntegrityError
+from odoo import api, exceptions, fields, models, _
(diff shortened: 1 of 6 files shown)


odoo/addons/mozaik_abstract_model/models/mozaik_abstract_model.py
```diff
@@ -87,6 +96,43 @@ def init(self):
...
+        Load error messages depending on index name and raise the
+        ValidationError with the correct message
+        :param exception: Exception object
+        :return:
+        """
+        messages = self._get_exception_messages()
+        message = messages.get(exception.diag.constraint_name)
+        if message:
+            raise exceptions.ValidationError(message)
+        # If we don't find a message, re-raise the exception
```
Olivier-LAURENT: Why the same ? It will produce the same effect, i.e. a bug report to the user !
Is it possible to raise a `ValidationError` with the original exception text ?
Otherwise we have to write a `_get_exception_messages(`) method in all model issued from `mozaik.abstract.model`.

Olivier-LAURENT:
In addition, a rebase is now needed.

PR #123, opened 2018-09-25
[CHG] remove bridge_field from distribution list and useless _get_computed_targets() method

Olivier-LAURENT:
In addition:
* virtual models are no longer auto_install
* complete list of valid models for each installed virtual model
* an order by into OVER() clause of virtual models are mandatory otherwise search() and read() results may be incompatible
* remove safe_mode parameter of _get_target_from_distribution_list()
* add a not store computed field result_id for each virtual model to translate common_id to a real id

### DIFF
--- a/odoo/addons/mozaik_communication/models/abstract_virtual_model.py
+++ b/odoo/addons/mozaik_communication/models/abstract_virtual_model.py
@@ -18,6 +18,11 @@ class AbstractVirtualModel(models.AbstractModel):
     common_id = fields.Char(
         string="Common ID",
     )
+    result_id = fields.Many2one(
+        comodel_name='virtual.target',
+        string='Result',
+        compute='_compute_result_id',
+    )
     int_instance_id = fields.Many2one(
         comodel_name='int.instance',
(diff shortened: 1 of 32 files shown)


odoo/addons/mozaik_communication/models/abstract_virtual_model.py
```diff
@@ -74,6 +79,19 @@ class AbstractVirtualModel(models.AbstractModel):
...
+    def _compute_result_id(self):
+        """
+        Compute result id based on common_id field
+        """
+        common_ids = self.mapped('common_id')
+        vts = self.env['virtual.target'].search(
+            [('common_id', 'in', common_ids)])
+        ids = {vt.common_id: vt.id for vt in vts}
+        for record in self:
+            record.result_id = ids[record.common_id]
```
acsonefho: In case of the `common_id` is not into your result (`ids`), you'll have an `Exception`.
You should replace this line by something like this:
`record.result_id = ids.get(record.common_id, self.env['virtual.target'].browse())`

PR #124, opened 2018-10-09
[ADD] Add ir_rule_child_of, addon to change the behavior of the child…

RobinetDenisAcsone:
… of in the ir.rule (return False in case of empty list)

### DIFF
--- a/odoo/addons/ir_rule_child_of/models/ir_rule.py
+++ b/odoo/addons/ir_rule_child_of/models/ir_rule.py
@@ -1,0 +1,31 @@
+# Copyright 2018 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, models, tools
+
+
+class IrRule(models.Model):
+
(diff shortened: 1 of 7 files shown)


odoo/addons/ir_rule_child_of/models/ir_rule.py
```diff
@@ -0,0 +1,31 @@
...
+        if dom:
+            dom = isinstance(dom, list) and dom or list(dom)
+            for ind, d in enumerate(dom):
+                if not isinstance(d, str) and len(d) == 3:
+                    if d[1] == 'child_of' and not d[2]:
+                        dom[ind] = (0, '=', 1)
+        return dom
+
+    @api.model
+    def clear_cache(self):
```
Olivier-LAURENT: Deprecated method, use  `clear_cacheS()` instead

Are you sure this inheriting is still needed ?

Olivier-LAURENT (changes requested):
In addition add this module in dependencies of mozaik addon

PR #140, opened 2019-01-17
[FIX] Performances and fixes changes

RobinetDenisAcsone:
* At the statement creation, search the partner based on the reference
* performance improvement in the renew wizard (force_lines
* log and performance for mass renew (split all the membership into several chunk)

### DIFF
--- a/odoo/addons/mozaik_account/models/account_bank_statement_line.py
+++ b/odoo/addons/mozaik_account/models/account_bank_statement_line.py
@@ -12,6 +12,14 @@ class AccountBankStatementLine(models.Model):
     _inherit = 'account.bank.statement.line'
 
     @api.model
+    def create(self, vals):
+        if not vals.get("partner_id"):
+            mode, partner = self._get_info_from_reference(vals.get("name"))
+            if partner:
+                vals["partner_id"] = partner.id
+        return super().create(vals)
+
+    @api.model
     def _get_models(self):
         return {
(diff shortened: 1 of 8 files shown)


odoo/addons/mozaik_account/models/account_bank_statement_line.py
```diff
@@ -11,6 +11,14 @@ class AccountBankStatementLine(models.Model):
 
     _inherit = 'account.bank.statement.line'
 
+    @api.model
+    def create(self, vals):
+        if not vals.get("partner_id"):
+            mode, partner = self._get_info_from_reference(vals.get("name"))
```
Olivier-LAURENT: `mode` is not used here, use a double-underscore instead

odoo/addons/mozaik_membership/models/membership_line.py
```diff
@@ -591,11 +605,29 @@ def _launch_renew(self, date_from=False):
...
         - Renew them
         :param date_from: str/date
-        :return: membership.line recordset
         """
-        lines = self._get_lines_to_close_renew()._close(date_to=date_from)
+        close_lines = self._get_lines_to_close_renew()
+
+        last_i = 0
+        step = int(self.env['ir.config_parameter'].get_param(
+            'membership.number_slice_renew', default='300'))
```
Olivier-LAURENT: `membership.renewal_slice_size`

Olivier-LAURENT (changes requested):
Just 2 details

PR #144, opened 2019-01-28
[CHG] virtual model: "store" the relation between the virtual model a…

RobinetDenisAcsone:
…nd virtual.target for optimisations


odoo/addons/distribution_list/models/distribution_list_line.py
```diff
@@ -236,7 +236,9 @@ def _get_target_recordset(self):
                 if field_name == 'id':
                     targets |= results
                 else:
-                    targets |= results.mapped(field_name)
+                    # to speedup use read and not mapped
+                    ids = set(r[field_name][0] for r in results.read([field_name]))
```
Olivier-LAURENT: It's better here to only get `r[field_name]` but forcing a `read(..., load='_classic_write')`

odoo/addons/mozaik_communication/models/virtual_target.py
```diff
@@ -12,6 +12,9 @@ class VirtualTarget(models.Model):
     ]
     _auto = False
 
+    result_id = fields.Many2one(
+        compute='_compute_result_id',
```
Olivier-LAURENT: Is this computed method still useful here ?

Olivier-LAURENT (changes requested):

PR #150, opened 2019-02-04
[FIX] fix virtual model for membership and export csv

RobinetDenisAcsone:


Olivier-LAURENT (changes requested):
The csv header is too long now

PR #164, opened 2019-05-20
[CHG][refs #34634] When the main address change, correctly manage the membership line

RobinetDenisAcsone:


odoo/addons/mozaik_account/wizards/partner_change_instance.py
```diff
@@ -24,9 +24,13 @@ def _create_new_membership(self, line, product):
                 product, line.partner_id, self.new_instance_id)
             ref = line._generate_membership_reference(
                 line.partner_id, self.new_instance_id)
+            reference, price = self.env["membership.line"]._prepare_custom_renew(
+                ref, price)
+            if not reference:
```
Olivier-LAURENT: Probably, just the case where reference is none must be checked here.

Olivier-LAURENT (changes requested):

PR #171, opened 2019-10-29
[CHG][refs #37324] Allow to reconcile on the past year based on a config parameter

RobinetDenisAcsone:


odoo/addons/mozaik_account/data/ir_config_parameter.xml
```diff
@@ -6,4 +6,8 @@
         <field name="key">membership.renew_must_paid</field>
         <field name="value">1</field>
     </record>
+    <record id="membership_renew_past_date" model="ir.config_parameter">
+        <field name="key">membership.renew.past.date</field>
```
Olivier-LAURENT: param name out-of-context, better to define `membership.allow.reconcile.last.year`

odoo/addons/mozaik_account/models/membership_line.py
```diff
@@ -41,19 +51,31 @@ def _get_membership_line_by_ref(self, reference, include_inactive=True):
...
         ]
-        if not include_inactive:
-            domain.append(('active', '=', True))
-        return self.search(domain, limit=1)
+        membership = self.search(domain, limit=1)
+        min_date_from = self._get_min_reconciliation_date()
+        if membership and membership.date_from < min_date_from:
+            if raise_exception:
+                raise UserError(_(
+                    "The membership you want to reconciled is too old"))
```
Olivier-LAURENT: ...to reconcile...

odoo/addons/mozaik_account/models/membership_line.py
```diff
@@ -29,8 +30,17 @@ class MembershipLine(models.Model):
...
+    def _get_min_reconciliation_date(self):
+        min_date_from = datetime.date.today().replace(day=1, month=1)
+        param = self.env["ir.config_parameter"].get_param(
+            "membership.renew.past.date", default="0")
+        if param != "0":
+            min_date_from = min_date_from.replace(year=min_date_from.year - 1)
+        return fields.Date.to_string(min_date_from)
+
+    @api.model
+    @tools.ormcache('reference', 'raise_exception')
```
Olivier-LAURENT, line 42: This cache has no longer perenity because its contents depends on a system parameter not included in the cache key.

It would be better to search membership by ref anyway, to put the result in the cache anyway and then either to ignore the result or to raise en exception.

odoo/addons/mozaik_account/models/membership_line.py
```diff
@@ -41,19 +51,31 @@ def _get_membership_line_by_ref(self, reference, include_inactive=True):
...
             lambda s: s.active and not s.move_id and not float_compare(
                 s.price, amount, precision_digits=precision))
         if len(memberships) > 1:
             raise UserError(_(
                 "More than one membership to reconcile are available"))
+        min_date_from = self._get_min_reconciliation_date()
+        if memberships and memberships.date_from < min_date_from:
+            if raise_exception:
+                raise UserError(_(
+                    "The membership you want to reconciled is too old"))
```
Olivier-LAURENT: ...to reconcile...

Olivier-LAURENT (changes requested):

odoo/addons/mozaik_account/models/membership_line.py
RobinetDenisAcsone: When writing/creating on ir.config_parameter, Odoo do a clear_caches

PR #176, opened 2019-11-08
[CHG][refs #37496] Add with_remuneration field

RobinetDenisAcsone:


Olivier-LAURENT:
👍 but can you also add filters on virtual model ? Just before gender filters for instance.

PR #197, opened 2020-06-25
[CHG] Correct double notes field in tree view; refs #40196

EcoloDevelopper:
close #40196


odoo_addons/mozaik_communication/distribution_list_view.xml
```diff
@@ -15,7 +15,6 @@
 
                 <xpath expr="//field[@name='alias_name']" position="replace">
                     <field name="code"/>
-                    <field name="note"/>
```
RobinetDenisAcsone: I think this line should be kept, since it's in the search.

The duplicated column in the issue come from the user action I think, so only the xpath below is necessary.

RobinetDenisAcsone (changes requested):

PR #204, opened 2020-12-16
46048 add interest on filter ara

EcoloDevelopper:


odoo_addons/mozaik_communication/models/virtual_partner_thesaurus_child_search.py
```diff
@@ -0,0 +1,75 @@
...
+    search_competencies_m2m_ids = fields.Many2many(
+        compute=lambda s: [],
+        comodel_name='thesaurus.term', store=False,
+        search="_search_competencies_m2m_ids")
+
+    @api.model
+    def _search_competencies_m2m_ids(self, operator, value):
+        if isinstance(value, str):
+            # the user doesn't select a item in the list
+            query = """SELECT id FROM thesaurus_term
```
RobinetDenisAcsone, line 29: The query below (when the user select a item) is recursive, but not this one. Is it intentional? It will not have the same result even if one therm is equals to the str value

EcoloDevelopper: Yes, this query is just to get an id of a thesaurus term with the string value. If there is an id in the first query I replace the function variable value with the id and then the recursive query is call. I just push the function with some comment. It will be easier for the future.

PR #207, opened 2021-02-03
48497 add derivate mandate

EcoloDevelopper:
close 48497


EcoloDevelopper:
Cancel pull request, needs info maybe change of logic for this request.

PR #230, opened 2021-07-30
[MIG] mozaik_partner_assembly: Migration to 14.0

qgroulard:

PR #232, opened 2021-08-05
[MIG] mozaik_structure: Migration to 14.0

qgroulard:
Needs https://github.com/mozaik-association/mozaik/pull/230
Includes migration of mozaik_abstract_model and inherit_abstract_view.

Migration of module 'disable_tracking_installation' (dependence of 'mozaik_abstract_model') is not included.
Would be preferable to drop it if not needed anymore.

No more group on 'archive' buttons.
Restrict usage of 'archive' button now requires JS code if I am not mistaken.


RobinetDenisAcsone:
What do you mean by "no menu included"? for structure, they are in mozaik_structure/views/structure_menu.xml  (a group is needed)

PR #233, opened 2021-08-20
Migrate mozaik_address and mozaik_address_drt

RobinetDenisAcsone:
mozaik_duplicate and mozaik_coordinate are also migrated but unused for now (since the behavior of mozaik_address have change)
The duplicate will only be applied in mozaik_person for the fields address, phones, email and name (mozaik_email and mozaik_phone shound not be migrated)


qgroulard (approved):
LGTM, code review.

Only commit history is a bit odd.
With commits for migration to v13 then modules version bumped to 14 in other modules' black commit.
And black changes in "not black commits".

PR #238, opened 2021-09-29
Mozaik distribution list ara

EcoloDevelopper:


distribution_list/views/assets.xml
```diff
@@ -2,9 +2,16 @@
...
+    <template
+        id="assets_backend"
+        name="filter_selection"
+        inherit_id="web.assets_backend"
+    >
         <xpath expr="." position="inside">
-            <script type="text/javascript" src="/distribution_list/static/src/js/filter_selection.js"/>
+            <script
+                type="text/javascript"
+                src="/distribution_list/static/src/js/filter_selection.js"
```
RobinetDenisAcsone: This js file was removed, so this assets.xml is unnecessary. You can remove it.

RobinetDenisAcsone (changes requested):
Can you also add the module in the mozaik_all depends list?

PR #240, opened 2021-10-13
[14.0] Migrate mozaik membership

RobinetDenisAcsone:


mozaik_membership/models/membership_line.py
```diff
@@ -620,17 +619,48 @@ def _launch_renew(self, date_from=False):
...
+        for i in range(step, len(close_lines), step):
+            close_lines[last_i:i]._close_and_former_member(date_from=date_from)
+            last_i = i
+        close_lines[last_i:]._close_and_former_member(date_from=date_from)
+
+    def _close_and_former_member(self, date_from=False):
+        # self.with_delay(
+        #     description="Former member %s memberships" % len(self),
+        #     channel="root.membership_close_and_former_member"
+        # )._job_close_and_former_member(date_from=date_from)
```
qgroulard: Do we need to keep this ?

qgroulard (commented):
LGTM, good work !

We should make sure to not forget to come back here after mozaik_involvement mig.

mozaik_membership/models/membership_line.py
RobinetDenisAcsone: I de-commented this and remove the direct call, it was for testing

PR #243, opened 2021-10-18
[ADD] Add mozaik_survey

marielejeune:


mozaik_survey/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2021 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Mozaik Survey",
+    "summary": """
+        This feature allows to make surveys""",
```
sbidoul, line 7: Arriving here by accident :)

Over time we noticed that it is very important to have a good (even if short) description of what a module does.

So we can easily know what a module is about, of course.

More importantly, because writing the description forces us to think about the content of the module, and detect when a module does unrelated things that need to go in separate modules.

So writing a `description` field or README.rst is a must.

PR #250, opened 2021-10-19
[ADD] mozaik_event_tickbox_question

marielejeune:


mozaik_event_tickbox_question/views/event_event.xml
```diff
@@ -8,6 +8,12 @@
         <field name="model">event.event</field>
         <field name="inherit_id" ref="website_event_questions.event_event_view_form" />
         <field name="arch" type="xml">
+            <field name="seats_expected" invisible="1" />
+            <xpath expr="//field[@name='event_type_id']" position="attributes">
+                <attribute name="attrs">
+                    {'readonly':[('seats_expected','>',0)]}
```
qgroulard, line 14: This seems a bit weak to me.
You only make the field readonly in the view, but it can be changed by the code.
I think a check in `write()` would be safer.

mozaik_event_tickbox_question/models/event_event.py
```diff
@@ -1,23 +1,31 @@
...
 
 
 class EventEvent(models.Model):
 
     _inherit = "event.event"
 
+    def write(self, vals):
+        """We add a check to prevent changing the type of an event
+        if there are already attendees."""
+        if self.seats_expected > 0:
```
qgroulard: for rec in self:
     if event_type_id in vals and vals["event_type_id"] != rec.event_type_id.id:

qgroulard (changes requested):

qgroulard (approved):

PR #254, opened 2021-10-20
[ADD] mozaik_mass_mailing_automation

marielejeune:
refs #53576

### DIFF
--- a/mozaik_mass_mailing_automation/models/mailing_mailing.py
+++ b/mozaik_mass_mailing_automation/models/mailing_mailing.py
@@ -1,0 +1,55 @@
+# Copyright 2021 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from datetime import datetime
+
+from dateutil.relativedelta import relativedelta
+
+from odoo import api, fields, models
+
+
+class MassMailing(models.Model):
+
+    _inherit = "mailing.mailing"
+
+    automation = fields.Boolean(string="Automation", default=False)
+    next_execution = fields.Datetime(
+        string="Next Execution of the Automation Process",
+        default=datetime.now(),
+        help="Date and time of the next execution. "
+        "New sendings will be planned every 24 hours.",
+    )
+    schedule_date = fields.Datetime(
+        string="Scheduled for",
+        tracking=True,
+        compute="_compute_schedule_date",
+        store=True,
+    )  # becomes a compute in case of automation
+
+    @api.depends("automation", "next_execution")
+    def _compute_schedule_date(self):
+        for record in self:
+            if record.automation:
+                record.schedule_date = record.next_execution
+            elif record.next_execution > datetime.now():
+                record.schedule_date = False
+
+    def _process_mass_mailing_queue(self):
+        mass_mailings = self.search(
+            [
+                ("automation", "=", True),
+                (
+                    "state",
+                    "in",
+                    ("in_queue", "sending", "done"),
+                ),  # in_queue is necessary to consider these mailings in for loop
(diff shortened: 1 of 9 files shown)


marielejeune:
@qgroulard review requested

mozaik_mass_mailing_automation/models/mailing_mailing.py
```diff
@@ -0,0 +1,55 @@
...
+    )
+    schedule_date = fields.Datetime(
+        string="Scheduled for",
+        tracking=True,
+        compute="_compute_schedule_date",
+        store=True,
+    )  # becomes a compute in case of automation
+
+    @api.depends("automation", "next_execution")
+    def _compute_schedule_date(self):
```
qgroulard, line 39: compute must always set a value, otherwise the field's value will be set to None.
If I understand correctly you need an extra
```
else:
    record.schedule_date = record.schedule_date
```

Please test this case (and other cases) as codecov suggests it.

mozaik_mass_mailing_automation/models/mailing_mailing.py
```diff
@@ -0,0 +1,55 @@
...
+        default=datetime.now(),
+        help="Date and time of the next execution. "
+        "New sendings will be planned every 24 hours.",
+    )
+    schedule_date = fields.Datetime(
+        string="Scheduled for",
+        tracking=True,
+        compute="_compute_schedule_date",
+        store=True,
+    )  # becomes a compute in case of automation
```
qgroulard: Confusing comment imo.
I think the `_compute_schedule_date()` method is simple enough to easily understand how this field works.

In the other hand, you may want to help users understand this field by adding a description.

mozaik_mass_mailing_automation/models/mailing_mailing.py
```diff
@@ -0,0 +1,55 @@
...
+                    ("in_queue", "sending", "done"),
+                ),  # in_queue is necessary to consider these mailings in for loop
+                ("schedule_date", "<", datetime.now()),
+            ]
+        )
+        mass_mailings.write({"state": "in_queue"})
+        super()._process_mass_mailing_queue()
+        for mailing in mass_mailings:
+            mailing.write(
+                {"next_execution": mailing.next_execution + relativedelta(days=1)}
```
qgroulard: What if mailing.next_execution = False ?

mozaik_mass_mailing_automation/models/mailing_mailing.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class MassMailing(models.Model):
+
+    _inherit = "mailing.mailing"
+
+    automation = fields.Boolean(string="Automation", default=False)
+    next_execution = fields.Datetime(
+        string="Next Execution of the Automation Process",
+        default=datetime.now(),
```
qgroulard: Default value seems dangerous to me.
It means that from now on every `mailing.mailing` will have a `next_execution` field filled.
I think scheduling a next_execution should always be done on purpose by a user.

Maybe I'm wrong ? You surely better know the use case than me.

PS: If you are aiming to ease encoding, you can add default in the view rather than in the model.

qgroulard (changes requested):

mozaik_mass_mailing_automation/models/mailing_mailing.py
marielejeune: This field won't be displayed.

mozaik_mass_mailing_automation/models/mailing_mailing.py
marielejeune: Can't happen (api.constraint) when mailing.automation = True

PR #258, opened 2021-10-22
[MIG] mozaik_account: Migration to 14.0

marielejeune:


sbidoul:
This is also a module that would be nice to split by feature.

PR #261, opened 2021-10-27
[UPD] mozaik_petition: read and write record rules

marielejeune:
refs #53354

@qgroulard review requested

### DIFF
--- a/mozaik_petition/models/petition_petition.py
+++ b/mozaik_petition/models/petition_petition.py
@@ -17,7 +17,13 @@ class PetitionPetition(models.Model):
     description = fields.Text(string="Description")
     note = fields.Text(string="Notes")
     image = fields.Binary(string="Image")
-    is_private = fields.Boolean(string="Is private")
+    visible_on_website = fields.Boolean(string="Visible on website", default=True)
+    is_private = fields.Boolean(
+        string="Is private",
+        help="If ticked, only members of authorized internal "
+        "instances have access to the petition.",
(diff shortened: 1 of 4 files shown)


mozaik_petition/models/petition_petition.py
```diff
@@ -77,6 +80,11 @@ def write(self, vals):
                 )
         return super(PetitionPetition, self).write(vals)
 
+    @api.depends("user_id")
```
sbidoul: There might be a risk of database inconsistency if the user is modified to be on another int_instance_id, as this will not be recomputed.

mozaik_petition/security/rules_petition_petition.xml
```diff
@@ -3,26 +3,32 @@
      License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
 <odoo>
 
-    <record id="petition_petition_rule_write_user" model="ir.rule">
-        <field name="name">Petition: user write own petitions rule</field>
+    <record id="petition_petition_rule_read_user" model="ir.rule">
```
qgroulard: Confusing name, rules are named 'read' but also give perm_write.

mozaik_petition/models/petition_petition.py
marielejeune: int_instance_id is now set to be equal to the internal instance of the user creating the petition, and cannot be modified later on.

PR #265, opened 2021-10-28
[ADD] mozaik_survey_mandatory_questions

marielejeune:
refs #52579

@qgroulard review requested


qgroulard:
It would be better to add a boolean `is_mandatory` on survey questions and then search questions with `is_mandatory=True` when loading default value for `question_and_page_ids`.
This is much more flexible than hard-coded questions.

marielejeune:

PR #268, opened 2021-10-29
[UPD] mozaik_event_is_private: adding record rules

marielejeune:
refs #54183

@qgroulard review requested

### DIFF
--- a/mozaik_event_is_private/models/event_event.py
+++ b/mozaik_event_is_private/models/event_event.py
@@ -8,4 +8,17 @@ class EventEvent(models.Model):
 
     _inherit = "event.event"
 
-    is_private = fields.Boolean(string="Is Private", default=False)
+    visible_on_website = fields.Boolean(string="Visible on website", default=True)
+    is_private = fields.Boolean(
+        string="Is private",
+        help="If ticked, only members of authorized internal "
+        "instances have access to the event.",
+        default=False,
+    )
+
+    int_instance_id = fields.Many2one(
+        "int.instance",
+        string="Internal instance",
+        help="Internal instance of the event",
+        default=lambda self: self.env.user.partner_id.int_instance_id,
+    )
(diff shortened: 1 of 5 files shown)


mozaik_event_is_private/models/event_event.py
```diff
@@ -8,4 +8,17 @@ class EventEvent(models.Model):
 
     _inherit = "event.event"
 
-    is_private = fields.Boolean(string="Is Private", default=False)
+    visible_on_website = fields.Boolean(string="Visible on website", default=True)
```
qgroulard, line 11: What does this field do in this module ?
Is it 'is_private' related in any way ?

mozaik_event_is_private/views/event_event.xml
```diff
@@ -6,10 +6,21 @@
     <record model="ir.ui.view" id="event_event_form_view">
         <field name="name">event.event.form (in mozaik_event_is_private)</field>
         <field name="model">event.event</field>
-        <field name="inherit_id" ref="event.view_event_form" />
+        <field
+            name="inherit_id"
+            ref="mozaik_event_publish_date.event_event_form_view"
```
qgroulard: Then your module needs to depend on `mozaik_event_publish_date` but I don't think we want that.
You should place your fields based on the base 'event' view.

mozaik_event_is_private/models/event_event.py
marielejeune: This field was present but unused. It's name is changed since we needed another field named "is_private"

mozaik_event_is_private/views/event_event.xml
marielejeune: Done

marielejeune:

mozaik_event_is_private/models/event_event.py
qgroulard: @marielejeune 
Keep in mind that you can override one (or some) field's parameter(s) without redefining all of them.
In your case you only need `visible_on_website = fields.Boolean(string="Visible on website")`.
Redefining other parameters is dangerous, the base module may change but those changes would be overridden by yours.

PR #269, opened 2021-10-29
[ADD] mozaik_survey_is_private: adding record rules

marielejeune:
refs #54098

@qgroulard review requested


mozaik_all/migrations/14.0.1.0.1/pre-migration.py
```diff
@@ -0,0 +1,8 @@
+# Copyright 2018 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from mozaik_survey_is_private.hook import pre_init_hook
+
+
+def migrate(cr, version):
```
qgroulard: pre_init_hook will be triggered when the module is installed, you don't need a migration script in mozaik_all for that

mozaik_survey_is_private/hook.py
```diff
@@ -0,0 +1,20 @@
...
+    env = odoo.api.Environment(cr, odoo.SUPERUSER_ID, {})
+    xml_ids = [
+        "survey.survey_survey_rule_survey_user_read",
+        "survey.survey_question_rule_survey_user_read",
+        "survey.survey_question_answer_rule_survey_user_read",
+        "survey.survey_user_input_rule_survey_user_read",
+        "survey.survey_user_input_line_rule_survey_user_read",
+    ]
+
+    for xml_id in xml_ids:
```
qgroulard: Do you really need to deactivate these rules ?
Can't you just add yours on top of these ones ?

mozaik_all/migrations/14.0.1.0.1/pre-migration.py
marielejeune: Was removed

mozaik_survey_is_private/hook.py
marielejeune: New group "User: child int instances only" and associated record rules were created.

marielejeune:

PR #280, opened 2021-11-10
Update dotfiles

sbidoul:


mozaik_abstract_model/models/mozaik_abstract_model.py
```diff
@@ -137,7 +137,7 @@ def _display_error_message(self, exception):
             # original expression but with a ValidationError
             constraint_name = exception.diag.constraint_name
         except AttributeError:
-            raise ValidationError(str(exception))
+            raise ValidationError(str(exception))  # noqa: B904
         message = self._get_exception_messages().get(constraint_name, str(exception))
         raise ValidationError(message)
```
sbidoul, line 142: Here, instead of raising the exception, this method should return the computed message, and let the caller raise the exception, so the stack trace remains easier to understand. @qgroulard @marielejeune @RobinetDenisAcsone

sbidoul:
FYI, this was done (mostly) by running `pre-commit autoupdate` to get the latest version of all linters.

Since this is not OCA where we risk breaking 200 repos when we bump linter versions, here we can afford to stay on latest versions.

PR #317, opened 2021-12-08
[ADD] REST APIs

qgroulard:


mozaik_membership_rest/services/membership_request.py
```diff
@@ -113,7 +113,7 @@ def membership_request(
         self, membership_request: MembershipRequest
     ) -> MembershipRequestInfo:
         vals = self._validate_membership_request_input(membership_request)
-        mr = self.env["membership.request"].create(vals)
+        mr = self.env["membership.request"].with_context(mode="autoval").create(vals)
```
qgroulard: @marielejeune  Why do you need the context key if you call validate_request() right after ?

mozaik_petition_rest/services/petition.py
```diff
@@ -89,9 +89,7 @@ def register_answer(
                 )
             )
         registration_values["registration_answer_ids"] = answers
-        petition_registration = (
-            self.env["petition.registration"]
-            .with_context(autoval=True)
-            .create(registration_values)
+        petition_registration = self.env["petition.registration"].create(
```
qgroulard, line 105: @marielejeune And why don't you need it here anymore ?

marielejeune:
@qgroulard review requested for the last commit please

mozaik_partner_rest/services/partner.py
```diff
@@ -70,3 +77,24 @@ def _prepare_update_values(self, values, partner):
                 )
             res["address_address_id"] = address.id
         return res
+
+    @restapi.method(
+        routes=[(["/<int:_id>/get_allowed_ama_objects"], "POST")],
```
marielejeune: This service should be "GET", but GET services do not support AMAObjectInfoList as input parameter

mozaik_partner_rest/services/partner.py
```diff
@@ -70,3 +77,24 @@ def _prepare_update_values(self, values, partner):
...
+        input_param=PydanticModel(AMAObjectInfoList),
+        output_param=PydanticModelList(AMAObjectInfo),
+    )
+    def get_allowed_ama_objects(
+        self, _id: int, ama_objects: AMAObjectInfoList
+    ) -> List[AMAObjectInfo]:
+        res: List[AMAObjectInfo] = []
+        for ama_object in ama_objects.ama_objects:
+            if ama_object.model in ["event", "petition", "survey"]:
+                model = ama_object.model + "." + ama_object.model
```
qgroulard, line 92: We should either get the complete model name in ama_object.model or add a dict for translation like "event": "event.event",...
But model_name + "." + model_name is too reliant on this specific case.

qgroulard: We should either get the complete model name in ama_object.model or add a dict for translation like "event": "event.event",...
But model_name + "." + model_name is too reliant on this specific case.

mozaik_partner_rest/services/partner.py
```diff
@@ -70,3 +77,24 @@ def _prepare_update_values(self, values, partner):
...
+                found_object = self.env[model].browse(ama_object.id)
+                allowed_partners = self.env["res.partner"].search(
+                    safe_eval(found_object.domain)
```
qgroulard: browse() does not guarantee that the record exists, you would get an error here if not.

mozaik_partner_rest/services/partner.py
qgroulard: I believe we could have 3 routes, one for each model.
Then we would take a list of ids instead of a list of AMAObjects.

qgroulard (commented):
@marielejeune 
I think this whole 'ama domain' thing is really specific to CDH.
We should extract this in a specific module.
I don't mind having this module in mozaik (this could be useful for other projects) but I think we should be able to install mozaik_partner_rest without pulling this part.

Plus, domain and domain_is_set fields make no sense without this REST service so they should be together in their own module.

mozaik_partner_rest/services/partner.py
qgroulard: Note that this comment makes no more sense if you decide to got for 3 routes instead of 1.

marielejeune:
Commits concerning domains on AMA objects were removed.

mozaik_membership_rest/services/service.py
```diff
@@ -15,7 +15,9 @@ class BaseMembershipRequestService(AbstractComponent):
 
     def _get(self, _id):
         domain = [("id", "=", _id)]
-        record = self.env[self._expose_model].search(domain)
```
qgroulard: What is the use case ?
I am not sure we always want to get archived membership requests.
I'd rather add a a parameter.

mozaik_membership_rest/services/membership_request.py
```diff
@@ -145,4 +146,12 @@ def membership_request(
...
-        return MembershipRequestInfo.from_orm(mr)
+        mr_info = MembershipRequestInfo.from_orm(mr)
+        if mr.state == "validate":
+            # Add active membership line info
+            active_ml = mr.partner_id.membership_line_ids.filtered(lambda ml: ml.active)
+            if active_ml:
+                active_ml_info = MembershipLineInfo.from_orm(active_ml)
+                mr_info.active_membership_line_info = active_ml_info
+
+        return mr_info
```
qgroulard: Why do you need this ? Membership_line_ids are already exposed on partners.

qgroulard (commented):
@marielejeune

mozaik_membership_rest/services/service.py
marielejeune: @qgroulard all membership request that were already validated are archived, hence we cannot read them anymore using this get.

qgroulard: Ok.
One must be careful when using _get() then, but I guess it is fair in such a case.

marielejeune:
Tests are failing because last commit depends on https://github.com/mozaik-association/mozaik/pull/759 that will be merged soon.

mozaik_involvement_rest/services/involvement.py
```diff
@@ -0,0 +1,83 @@
...
+        routes=[(["/", "/search"], "GET")],
+        input_param=PydanticModel(InvolvementSearchFilter),
+        output_param=PydanticModelList(InvolvementInfo),
+    )
+    def search(
+        self, involvement_search_filter: InvolvementSearchFilter
+    ) -> List[InvolvementInfo]:
+        domain = self._get_search_domain(involvement_search_filter)
+        res: List[InvolvementInfo] = []
+        for ic in self.env["partner.involvement"].sudo().search(domain):
```
qgroulard: ic ?

marielejeune: Copy-paste ;-) Fixed

mozaik_country_rest/services/country.py
```diff
@@ -44,6 +44,6 @@ def _get_search_domain(self, filters):
     def search(self, country_search_filter: CountrySearchFilter) -> List[CountryInfo]:
         domain = self._get_search_domain(country_search_filter)
         res: List[CountryInfo] = []
-        for c in self.env["res.country"].sudo().search(domain):
```
marielejeune: @qgroulard please have a look at this change.
It makes no sense to me to add a `sudo()` here as there exists a global read access on `res.country`...

marielejeune:
I've created another PR for adding tests/fixes and trying to merge this stuff.
This PR is intended to be the one running on our clients database.

See https://github.com/mozaik-association/mozaik/pull/782 for merging attempts.

PR #318, opened 2021-12-08
[MIG] ecolo_mandate

qgroulard:


qgroulard:
@RobinetDenisAcsone Here is my work on ecolo_mandate
In first commit I tried to extract is_important into a new addon.
In second commit I directly added the logic in existing addons.

Both approaches fail for the same reason.

PR #322, opened 2021-12-08
[FIX] mozaik_event_tickbox_question: constraint on event_registration…

marielejeune:
… for mandatory questions


marielejeune:
@qgroulard review requested

PR #323, opened 2021-12-08
[FIX] mozaik_petition: constraint on petition_registration for mandat…

marielejeune:
…ory questions


marielejeune:
@qgroulard review requested

PR #327, opened 2021-12-10
Remove interest from event, petition, survey

marielejeune:


marielejeune:
@qgroulard review requested

PR #340, opened 2021-12-17
[FIX] distribution_list: mass mailings are sent to the right partners

marielejeune:
refs #55756


marielejeune:

PR #359, opened 2022-01-13
[ADD] buttons on partner to add new mandates with default_partner_id

marielejeune:
refs #56330

### DIFF
--- a/mozaik_committee/models/abstract_selection_committee.py
+++ b/mozaik_committee/models/abstract_selection_committee.py
@@ -152,6 +152,17 @@ class AbstractSelectionCommittee(models.Model):
 
     # view methods: onchange, button
 
+    def add_candidature_action(self):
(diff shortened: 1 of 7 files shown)


mozaik_committee/models/ext_selection_committee.py
```diff
@@ -40,3 +40,8 @@ class ExtSelectionCommittee(models.Model):
         string="Members",
         domain=[("is_company", "=", False)],
     )
+
+    def add_ext_candidature_action(self):
+        return self.add_candidature_action(
+            "Add a new External Candidature", "ext.candidature"
```
qgroulard: Not translatable without `_()`.

qgroulard (commented):
You could have a single function by using generic names such as "Add a new Candidature/Mandate" and getting the res_model from context.
It would be cleaner imo.

PR #363, opened 2022-01-13
[ADD] auth_api_key

qgroulard:


oca_dependencies.txt
```diff
@@ -21,3 +21,4 @@ queue https://github.com/mozaik-association/queue 11.0-mrm_master
 partner-contact https://github.com/mozaik-association/partner-contact 11.0-mrm_master
 social https://github.com/mozaik-association/social 11.0-mrm_master
 rest-framework https://github.com/acsone/rest-framework 14.0-mig-graphql_base-qgr
+server-auth
```
sbidoul, line 24: You can remove this oca_dependencies.txt, it is not used by the GitHub action tests.
Also, pointing to 11.0 branch on 14.0 would not work if it was used ;)

PR #393, opened 2022-01-26
Res partner features mle

marielejeune:


mozaik_partner_website/models/res_partner.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2022 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class ResPartner(models.Model):
```
qgroulard: You don't need an empty class.

mozaik_partner_website/views/res_partner.xml
```diff
@@ -0,0 +1,21 @@
...
+
+    <record model="ir.ui.view" id="res_partner_form_view">
+        <field name="name">res.partner.form (in mozaik_partner_website)</field>
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="website_partner.view_partners_form_website" />
+        <field name="arch" type="xml">
+            <field name="is_published" position="attributes">
+                <attribute name="attrs">
+                    {'invisible': 1}
+                </attribute>
```
qgroulard:
```suggestion
                <attribute name="invisible">1</attribute>
```
Avoid to use attrs when possible, this is much harder to extend.
Because attrs is char if you want to extend the 'dictionary' you need to completely overwrite it.

qgroulard (changes requested):

PR #396, opened 2022-01-28
14.0 mozaik survey rest v2 mle

marielejeune:


qgroulard:
Added to https://github.com/mozaik-association/mozaik/pull/317

PR #404, opened 2022-02-03
email and postal bounce counter

marielejeune:


mozaik_mass_mailing_bounce_counter/models/res_partner.py
```diff
@@ -0,0 +1,76 @@
...
+        """
+        Look at mass mailings sent since the last execution.
+        Update bounce counters:
+        if bounced email: increment the counter, write failure date
+        if no bounced email and counter > 0 : counter set back to 0.
+        """
+        # Last execution of the cron
+        last_execution = self.env.ref(
+            "mozaik_mass_mailing_bounce_counter.ir_cron_res_partner_email_bounced"
+        ).lastcall or datetime.now() - relativedelta(days=1)
```
qgroulard: It's weird to have cron logic here, you'd rather get last_execution as argument.

mozaik_mass_mailing_bounce_counter/models/res_partner.py
```diff
@@ -0,0 +1,74 @@
...
+    @api.model
+    def _update_bounce_counter(self, last_call):
+        """
+        Look at mass mailings sent since the last execution.
+        Update bounce counters:
+        if bounced email: increment the counter, write failure date
+        if no bounced email and counter > 0 : counter set back to 0.
+        """
+        # Last execution of the cron
+        last_execution = last_call or datetime.now() - relativedelta(days=1)
```
qgroulard: imo this should also go into the cron's code

PR #414, opened 2022-02-11
[UPD] mig script: setting email_bounced=0 by default on partners

marielejeune:


mozaik_all/migrations/14.0.1.1.4/pre-migrate.py
```diff
@@ -0,0 +1,16 @@
...
+from odoo import SUPERUSER_ID, api
+
+_logger = logging.getLogger(__name__)
+
+
+def migrate(cr, version):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    _logger.info("Setting email_bounced = 0 if None")
+
+    partners = env["res.partner"].search([("email_bounced", "=", None)])
```
qgroulard: No need for inactive partners ?
You could use SQL query to be sure to not miss any record.

PR #426, opened 2022-02-16
14.0 57144 always autovalidate mle

marielejeune:


mozaik_membership_request_autovalidate/models/membership_request.py
```diff
@@ -26,11 +56,14 @@ def _auto_validate(self, auto_val):
...
             self.phone,
         ]
-        nb_match_emails = 0
 
         # 26993/2.1.1 and 2.3.2 (first part)
         if auto_val and any(fields_tokeep_empty):
             auto_val = False
+            failure_reason = _(
+                "One of the following fields "
+                "is completed: street_man, number, box, mobile, phone"
```
qgroulard: You should get the list of field names from fields_tokeep_empty.
Otherwise one could forget to change the error message when changing the "tokeep_empty" list.

mozaik_membership_request_autovalidate/models/res_partner.py
```diff
@@ -0,0 +1,42 @@
...
+            "note": note,
+            "date_deadline": datetime.datetime.today() + datetime.timedelta(days=5),
+            "automated": True,
+            "res_model_id": self.env["ir.model"]
+            .search([("model", "=", "res.partner")])
+            .id,
+            "res_id": self.id,
+        }
+        control_user_id = int(
+            self.env["ir.config_parameter"].get_param("force_autoval_control_user")
```
qgroulard, line 34: Only one control user looks limiting to me...
Hopefully that will do the job for a phase 1 but this approach is not really scalable.

mozaik_event_membership_request_involvement/models/event_registration.py
```diff
@@ -10,6 +11,9 @@ class EventRegistration(models.Model):
 
     @api.model
     def create(self, vals):
+        # If associated_partner_id is not given, then partner_id cannot be given as well.
+        if "associated_partner_id" not in vals and "partner_id" in vals:
+            vals.pop("partner_id")
```
qgroulard, line 16: Then REST services will be broken ?

mozaik_event_membership_request_involvement/models/event_registration.py
```diff
@@ -21,7 +25,29 @@ def _create_membership_request_from_registration(self, vals):
         if request:
             request.write({"event_registration_id": self.id})
             self._add_involvements_to_membership_request(request)
-            request._auto_validate(self.event_id.auto_accept_membership)
+            failure_reason = request._auto_validate(
+                self.event_id.auto_accept_membership
+            )
+            force_autoval = self._context.get("force_autoval", False)
```
qgroulard, line 31: TO DO: adapt rest services.

mozaik_membership_request_autovalidate/models/membership_request.py
marielejeune: Done

mozaik_membership_request_autovalidate/models/res_partner.py
marielejeune: I was asked to do it like this for now. It will be improved later on.

marielejeune:
There are still some problems with partner_id on events: this field encodes the partner registering people on events, not the partners who will be registered. It implies that firstnames and lastnames can be modified in an unsafe way when forcing autovalidation.

PR #431, opened 2022-02-18
14.0 registered partner id mle

marielejeune:


mozaik_event_membership_request_involvement/models/event_registration.py
```diff
@@ -1,16 +1,42 @@
...
+                )
+
     @api.model
     def create(self, vals):
-        rec = super().create(vals)
+        registered_partner = False
+        if "associated_partner_id" in vals:
+            registered_partner = vals["associated_partner_id"]
+        rec = super(
+            EventRegistration, self.with_context({"mr_partner_id": registered_partner})
```
qgroulard: Why do you add the context here and not when calling _create_membership_request_from_registration ?

marielejeune: Done

PR #436, opened 2022-02-21
[ADD] mozaik_mandate_show_website

marielejeune:


mozaik_graphql/schema/types/mandate.py
```diff
@@ -21,6 +21,8 @@ class AbstractMandate(AbstractObject):
     start_date = graphene.DateTime(required=True)
     deadline_date = graphene.DateTime(required=True)
     end_date = graphene.DateTime()
+    no_show_on_website = graphene.Boolean()
```
qgroulard, line 24: add mozaik_mandate_show_website to mozaik_graphql dependencies

mozaik_graphql/schema/types/mandate.py
```diff
@@ -21,6 +21,8 @@ class AbstractMandate(AbstractObject):
...
+    partner_no_show_on_website = graphene.Boolean()
```
qgroulard: This is not really a mandate property... I don't really see the point to expose this here.
If they want to access this info they simply have to query {partner{no_show_mandates}}

PR #438, opened 2022-02-21
[UPD] mozaik_membership_request_from_registration: includes force autoval process

marielejeune:


marielejeune:
Purpose of this change: make a generic function that can be called in the membership_request REST service when auto-validation will be forced.

PR #446, opened 2022-02-25
[UPD] mozaik_mass_mailing_access_rights: add reader group

marielejeune:


mozaik_mass_mailing_access_rights/models/mailing_mailing.py
```diff
@@ -0,0 +1,16 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class MassMailing(models.Model):
+
+    _inherit = "mailing.mailing"
+
+    int_instance_ids = fields.Many2many(
```
qgroulard: Where is this field used ?

marielejeune: It will be used in cdh_security and ecolo_security

qgroulard: Then it should be in those modules imo.

PR #462, opened 2022-03-04
[ADD] mozaik_retrocession_mode

marielejeune:


mozaik_retrocession_mode/models/abstract_mandate.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2022 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+RETROCESSION_MODES_AVAILABLE = [
+    ("month", "Monthly"),
+    ("year", "Yearly"),
+    ("none", "None"),
```
qgroulard, line 9: Why a "none" value while you can simply leave the field empty ?

qgroulard (commented):
Having "none" in a selection values looks odd to me.
Is there any good reason why you would need that ?
Why is the field required as well ?

PR #474, opened 2022-03-10
[WIP] trying to fix the error on 'all involvements' button on res.partners

marielejeune:


marielejeune:
Bug case: clicking on button "All involvements" from a res.partner form view (defined in mozaik_involvement).
It seems that it is linked with module mozaik_involvement_followup since when uninstalling this module, there is no bug anymore.
With mozaik_involvement_followup, sometimes the bug doesn't appear anymore. Then, uninstalling and installing back mozaik_involvement_followup brings the bug back.

PR #483, opened 2022-03-17
[FIX] Address technical_name

qgroulard:
@RobinetDenisAcsone 
Does this feels safe to you ?
We have got an issue because `technical_name` is in the `unicity_keys`.
To compute the `technical_name`, only the `zip` is taken from the `city_id`.
But really often the same address can be written with 2 different cities but the same zip.
For instance:
Grand Place 1, 4800 Ensival
Grand Place 1, 4800 Verviers


RobinetDenisAcsone:
@qgroulard I don't see any problem, but there are one test failing

PR #486, opened 2022-03-23
[UPD] add bypass_ir_rule_read in view context for distribution.list 

marielejeune:
and partner.involvement.category

### DIFF
--- a/distribution_list/views/distribution_list.xml
+++ b/distribution_list/views/distribution_list.xml
@@ -135,6 +135,7 @@
         <field name="name">Distribution Lists</field>
         <field name="res_model">distribution.list</field>
         <field name="view_mode">tree,form</field>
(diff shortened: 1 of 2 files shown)


sbidoul:
This is certainly worth a comment explaining why we need to do this.

marielejeune:
@sbidoul mozaik module user_bypass_security allows to bypass ir.rules for reading on Many2manySudoRead fields. In v8 it added the context key 'bypass_ir_rule_read':'1'  in the context of the concerned field inside the view, but this process doesn't work anymore in v14, this context key is lost when reaching _apply_ir_rules method. After days of debugging the temporary solution is to add the key context to the whole view, since all many2many fields to res.users and res.partners on both models distribution.list and partner.involvement.category are concerned.
But we know that it isn't the best solution, hence another solution will be found asap.

PR #500, opened 2022-03-30
[RMV] user_bypass_security and modify get_partner_from 

marielejeune:


mozaik_communication/models/distribution_list.py
```diff
@@ -237,3 +248,25 @@ def write(self, vals):
...
+        partners = self.env["res.partner"].browse()
+        if len(self) != 1:
+            return partners
+        # first: the sender partner
+        partners |= self.partner_id
+        # than: the requestor user
+        if self.env.user.partner_id in self.res_partner_ids:
+            partners |= self.env.user.partner_id
+        elif self.env.user in self.res_users_ids:
+            partners |= self.env.user.partner_id
```
qgroulard:
```suggestion
        if self.env.user.partner_id in self.res_partner_ids or self.env.user in self.res_users_ids:
            partners |= self.env.user.partner_id
```

mozaik_communication/models/distribution_list.py
```diff
@@ -237,3 +248,25 @@ def write(self, vals):
...
         return res
+
+    def _get_partner_from(self):
+        """
+        Get partner from distribution list
+        :return: res.partner recordset
+        """
+        partners = self.env["res.partner"].browse()
+        if len(self) != 1:
+            return partners
```
qgroulard, line 259: `self.ensure_one()` ?

marielejeune: If len(self) != 1 (and it happens sometimes when opening the wizard) we don't want an error, we just want to return 0 partners

PR #530, opened 2022-04-26
[UPD] mozaik_event_tickbox_question, mozaik_petition:

marielejeune:
on questions, is_mandatory has to be false if question is not a tickbox

refs #59032


qgroulard (approved):
Keep in mind that onchange is only triggered from the view.
If question_type might be modified from code, you'd want to do this in create and write functions instead.
If not, you can merge as is.

PR #545, opened 2022-05-03
[FIX] mozaik_membership_request: interests were erased when pre_processing

marielejeune:
When creating a membership request and pre-processing the data, interest and competencies were not taken into account.

### DIFF
--- a/mozaik_membership_request/models/membership_request.py
+++ b/mozaik_membership_request/models/membership_request.py
@@ -629,7 +629,10 @@ class MembershipRequest(models.Model):
         res = self._onchange_partner_id_vals(
             is_company, request_type, partner_id, technical_name
         )
+        old_interests = res.pop("interest_ids", [])
+        old_competencies = res.pop("competency_ids", [])
         vals.update(res)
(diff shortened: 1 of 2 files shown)


mozaik_membership_request/models/membership_request.py
```diff
@@ -658,6 +661,29 @@ def _pre_process(self, vals):
 
         return vals
 
+    def _merge_interests_competencies(self, vals, old_interests, old_competencies):
+        """
+        vals is a dict that may already contain a command for interest_ids or competency_ids.
+        We want to integrate to this command, the previous interests and competencies.
+
+        old_interests is either [] or [(6, False, [list_ids])]
+        old_competencies is either [] or [(6, False, [list_ids])]
```
RobinetDenisAcsone, line 670: Are you sure that it's always a 6? and not another magic number?

Otherwise LGTM

marielejeune: Yes because it directly comes from onchange_partner_id_vals method

PR #548, opened 2022-05-03
14.0 59002 mozaik event barcode mle

marielejeune:


marielejeune:
@qgroulard here is an improvement of mozaik_event_barcode. 
Could you please review when you'll have time. Thanks

PR #549, opened 2022-05-04
14.0 59253 mozaik mass mailing immediate sending mle

marielejeune:


mozaik_mass_mailing_immediate_sending/migrations/14.0.1.0.1/post-migrate.py
```diff
@@ -0,0 +1,19 @@
...
+def migrate(cr, version):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    _logger.info("'Email Marketing: process queue' cron set to run every 5 minutes.")
+    cron_id = env.ref("mass_mailing.ir_cron_mass_mailing_queue").id
+    query = """
+    UPDATE ir_cron
+    SET interval_number = 5
+    WHERE id = %(id)s;
+    """
+    cr.execute(query, {"id": cron_id})
```
qgroulard, line 19: I don't see the point of doing this in a migration script.
Other customers could ask for a different interval, thus it is better to configure this via the interface.

PR #569, opened 2022-05-11
[FIX] virtual models: search on m2m fields computed non stored

marielejeune:
refs #59449


marielejeune:
In virtual models, searches on interest and competencies were not performed correctly. 
An error was raised since these m2m fields were computed non stored, without a search method.

PR #575, opened 2022-05-13
[FIX] Partner is_address_duplicate_allowed migration

qgroulard:
`is_address_duplicate_allowed_compute` and `is_address_duplicate_allowed` can not be computed properly if not initialized.

This is due to the fact that `field_allowed` is set to False if not True on both duplicates: https://github.com/mozaik-association/mozaik/blob/00262e81103857b13f7fa73c6234ccf6b72bff9d/mozaik_duplicate/models/abstract_duplicate.py#L254

### DIFF
--- a/mozaik_person/migrations/14.0.1.0.0/post-migration.py
+++ b/mozaik_person/migrations/14.0.1.0.0/post-migration.py
@@ -9,3 +9,14 @@ def migrate(cr, version):
     cr.execute(
         "UPDATE res_partner SET identifier = identifier_moved0",
     )
+
+    _logger.info("Init is_address_duplicate_allowed from co_residency")
+
+    cr.execute(
+        """
+        UPDATE res_partner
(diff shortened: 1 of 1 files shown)


mozaik_person/migrations/14.0.1.0.0/post-migration.py
```diff
@@ -9,3 +9,14 @@ def migrate(cr, version):
     cr.execute(
         "UPDATE res_partner SET identifier = identifier_moved0",
     )
+
+    _logger.info("Init is_address_duplicate_allowed from co_residency")
+
+    cr.execute(
```
RobinetDenisAcsone, line 15: This will run? The addon version should be changed and the code put in another migration file I think

qgroulard: I ran it manually in prod.
I'va added this here for future migrations (MR) and assigned to you in order to let you know that it was missing from your dev.

PR #616, opened 2022-07-04
[ADD] mozaik_mass_mailing_sending_cron

marielejeune:
refs #60581


qgroulard (commented):
We should clearly state somewhere why we are doing this.
We'll certainly be able to drop this module in a future Odoo version.

PR #619, opened 2022-07-12
[UPD] mozaik_membership_request: cron for member candidates not paying

marielejeune:
refs #60533

### DIFF
--- a/mozaik_membership_request/models/membership_line.py
+++ b/mozaik_membership_request/models/membership_line.py
@@ -1,9 +1,12 @@
 # Copyright 2021 ACSONE SA/NV
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+import logging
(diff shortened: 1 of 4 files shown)


mozaik_membership_request/models/membership_line.py
```diff
@@ -48,12 +51,51 @@ def create_following_membership(self, status, amount=False):
...
+        supporter_state = self.env["membership.state"].search(
+            [("code", "=", "supporter")]
+        )
+        for m in memberships:
+            _logger.info(
+                "Partner with id %(partner_id)s is member candidate "
+                "since more than %(days)s days "
+                "and didn't pay: he becomes a supporter."
+                % {"days": days, "partner_id": m.partner_id.id}
+            )
```
qgroulard: What is the purpose of this ?
This does not look like a log message, more like a chatter message.

PR #622, opened 2022-08-05
[ADD] mozaik_single_membership: add previous_state_id on membership lines

marielejeune:
refs #61461

### DIFF
--- a/mozaik_single_membership/models/membership_line.py
+++ b/mozaik_single_membership/models/membership_line.py
@@ -1,0 +1,34 @@
+# Copyright 2022 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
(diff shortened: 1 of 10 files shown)


mozaik_single_membership/models/membership_line.py
```diff
@@ -0,0 +1,34 @@
...
+
+    previous_state_id = fields.Many2one(
+        comodel_name="membership.state", string="Previous State"
+    )
+
+    def _update_previous_state(self):
+        self.ensure_one()
+        partner = self.partner_id
+        previous_line = first(
+            self.env["membership.line"].search(
```
RobinetDenisAcsone: Maybe add a limit=1 in the search, it may improve a little the performance for the mass renew

Also, perhaps modifying the vals would be better than doing a write (get the partner from vals, and add the previous_state_id in the vals before the super create call)

PR #633, opened 2022-09-20
[ADD] res.partner: mandate_phone field

marielejeune:
refs #63158


marielejeune:
@qgroulard I would like to rename this module: mozaik_mandate_email -> mozaik_mandate_email_phone. Should I do something special for the already existing res.groups record since its external ID will change?

qgroulard:
@marielejeune openupgrade lib has a method [update_module_names](https://github.com/OCA/openupgradelib/blob/f97bf37a63b0f44b1803293a2a98b91c47932593/openupgradelib/openupgrade.py#L1529)

qgroulard (commented):
imo it is worth also changing the res.group name since it now allows to see more than emails.

PR #636, opened 2022-09-21
[CHG] mozaik_all: res.partner form view: add social_media page for companies

marielejeune:
refs #62388


qgroulard (commented):
Shouldn't you remove the fields from the other page then ?

marielejeune:
> Shouldn't you remove the fields from the other page then ?

@qgroulard No because this page is only visible for companies, while the other one is only visible if you're not a company.

PR #666, opened 2022-12-01
[CHG] Complete re-writing of mozaik membership request pre-process

marielejeune:
Significant changes to mozaik membership requests pre-process and validation:

1. Use the same pre-process method for MR coming from an AMA object (event, petition, survey) and for MR coming from REST services.

3. Fix some bugs coming at validation of the request, linked to addresses and instances, and modify the behavior (in accordance with the clients) to be more consistent.
* Concerning the instances:
o     The instance is modified if and only if the address is.
o If a force instance is given, it is always taken into account.
* Concerning the addresses: 
o A complete address (with a street) will always be written on the partner.
o A partial address (with city but without street) will be written on the partner only if we had no address before.

3. Fix some other bugs, such as:
* When opening all MR of a partner, do not pass `active_test=False` in context, as we do not want to keep it in all further actions, on all models.
* When changing partner's instance, the new membership line must be marked as paid if the previous one, still active, was.
* When trying to recognize an existing partner, search directly on `res.partner` and not on `virtual.custom.partner`anymore (makes no sense as it excluded partners having no email and no postal address, which is unwanted)

### DIFF
--- a/mozaik_address_local_street/models/address_address.py
+++ b/mozaik_address_local_street/models/address_address.py
@@ -83,3 +83,9 @@ class AddressAddress(models.Model):
                 record.street_man = record.address_local_street_id.local_street
                 record.select_alternative_address_local_street = False
             record.street_man = False
+
+    @api.depends("street_man", "street2", "address_local_street_id")
+    def _compute_has_street(self):
+        super()._compute_has_street()
+        for address in self.filtered("address_local_street_id"):
+            address.has_street = True
--- a/mozaik_membership_request/models/membership_request.py
+++ b/mozaik_membership_request/models/membership_request.py
@@ -1343,41 +1343,17 @@ class MembershipRequest(models.Model):
         Make special combinations of domains to try to find
         a unique partner_id
         """
-        partner_obj = self.env["virtual.custom.partner"]
         partner_domains = self._get_partner_domains(recognizable_values)
-        partner_id = False
-        virtual_partner_id = self.persist_search(partner_obj, partner_domains)
-        # because this is not a real partner but a virtual partner
-        if virtual_partner_id:
-            partner_id = virtual_partner_id.partner_id
-        return partner_id
+        for domain in partner_domains:
+            try:
+                safe_domain = safe_eval.safe_eval(domain)
+            except Exception as e:
+                raise ValidationError(_("Invalid Data")) from e
+            res_ids = self.env["res.partner"].search(safe_domain)
+            if len(res_ids) == 1:
+                return res_ids
 
-    def persist_search(self, model_obj, domains):
-        """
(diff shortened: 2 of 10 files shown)


mozaik_membership_request/models/membership_request.py
```diff
@@ -656,103 +656,254 @@ def _detect_changes(self):
...
-                and not (address_local_street_id or street_man)
-                and (partner_address_id.zip == zipcode)
-            ):
-                address_id = False
-
         return address_id
 
-    @api.model
-    def _pre_process(self, vals):
+    def _find_input_partner(self, vals):
```
marielejeune, line 658: This method comes from `mozaik_membership_request_from_registration`, nothing new

mozaik_membership_request/models/membership_request.py
```diff
@@ -1343,14 +1343,9 @@ def get_partner_id(self, recognizable_values):
...
         partner_domains = self._get_partner_domains(recognizable_values)
-        partner_id = False
-        virtual_partner_id = self.persist_search(partner_obj, partner_domains)
-        # because this is not a real partner but a virtual partner
-        if virtual_partner_id:
-            partner_id = virtual_partner_id.partner_id
-        return partner_id
+        return self.persist_search(partner_obj, partner_domains)
 
     def persist_search(self, model_obj, domains):
```
qgroulard: Please get rid of this awful method as well :sob:

It is only called from `get_partner_id()` from what I see (I haven't check in each projects, only in mozaik).

mozaik_address_local_street/models/address_address.py
```diff
@@ -83,3 +83,9 @@ def _onchange_address_local_street_id(self):
                 record.street_man = record.address_local_street_id.local_street
                 record.select_alternative_address_local_street = False
             record.street_man = False
+
+    @api.depends("street_man", "street2", "address_local_street_id")
+    def _compute_has_street(self):
+        super()._compute_has_street()
+        for address in self.filtered(lambda ad: ad.address_local_street_id):
```
qgroulard: FYI, nitpicking you can leave it like that.
You don't need to use lambda to check for booleans.

`self.filtered("address_local_street_id")`

mozaik_membership_request/models/membership_request.py
```diff
@@ -823,11 +926,37 @@ def _pre_process(self, vals):
                 "country_id": country_id,
                 "zip_man": zip_man,
                 "city_man": city_man,
-                "technical_name": technical_name,
             }
         )
+        # We know check all other keys from input_vals:
```
qgroulard: Nit *now

RobinetDenisAcsone (approved):

PR #670, opened 2022-12-09
[IMP] mozaik_virtual_partner_mass_mailing: take translated version of mass mailing subject

marielejeune:


mozaik_virtual_partner_mass_mailing/models/virtual_partner_mass_mailing.py
```diff
@@ -96,10 +103,21 @@ def _get_where(self):
...
+            "AND p.active = TRUE "
+            "AND p.identifier IS NOT NULL "
+            "AND p.identifier != '0' "
+            "AND mm.active = TRUE "
+            "AND t.name='mailing.mailing,subject' "
+            "AND t.lang=%(user_lang)s "
         )
 
+    @api.model
+    def _get_query_parameters(self, parameter=False):
```
RobinetDenisAcsone, line 116: This is called on the update of the plugin (perhaps also on the startup of odoo) which means that the user is always sudo, and not the user that do the query at runtime.

RobinetDenisAcsone (changes requested):

PR #679, opened 2022-12-23
Make sure to alwayse call the membership validation in mozaik_members…

RobinetDenisAcsone:
…hip_payment to correctly mark the membership as paid if needed

### DIFF
--- a/mozaik_membership_payment/models/membership_request.py
+++ b/mozaik_membership_payment/models/membership_request.py
@@ -79,7 +79,7 @@ class MembershipRequest(models.Model):
         return [self.env.ref("mozaik_membership.former_member")]
 
     @api.model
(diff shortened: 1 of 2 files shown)


marielejeune:
@RobinetDenisAcsone is there any reason for _validate_request_membership and _validate_request_membership_with_checks to be decorated with api.model and take the MR as a parameter?

RobinetDenisAcsone:
@marielejeune I don't think, it could be an @api.mulit with self.ensure_one()

marielejeune:
@RobinetDenisAcsone in v14 api.multi is deprecated, so I'll just modify the methods and call self.ensure_one()

marielejeune:
Modification done here: https://github.com/mozaik-association/mozaik/pull/683

PR #698, opened 2023-01-20
[ADD] mozaik_email_lowered: mixin adding a new field 'lowered_email'

marielejeune:
This allows to transform searches with domain tuple ('email', '=ilike', value) into domain tuple ('lowered_email', '=', value) hence special characters such as % and _ are escaped.

refs #65665

### DIFF
--- a/mozaik_email_lowered/models/lowered_email_mixin.py
+++ b/mozaik_email_lowered/models/lowered_email_mixin.py
@@ -1,0 +1,41 @@
+# Copyright 2023 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+from odoo.addons.mozaik_tools.tools import format_email
+
+
+class MozaikLoweredEmailMixin(models.AbstractModel):
+
+    _name = "mozaik.lowered.email.mixin"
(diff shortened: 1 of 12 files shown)


marielejeune:
@qgroulard what do you think about this solution?

Remaining questions/comments:
* Is there a better solution to deal with `ilike` (not `=ilike`), for example in `res.partner` search view? Up to now the `_` char is not escaped in `ilike` case.
* Is the method efficient enough?
* Do you think we should add this mixin to other models (such as `event.registration`, `petition.registration` and probably some others)?

mozaik_email_lowered/models/lowered_email_mixin.py
```diff
@@ -0,0 +1,41 @@
...
+        new_args = []
+        for arg in args:
+            if (
+                isinstance(arg, (list, tuple))
+                and arg[0] == "email"
+                and arg[1] in ["ilike", "=ilike"]
+                and isinstance(arg[2], str)
+            ):
+                search_field_name = "lowered_email"
+                operator = "=" if arg[1] == "=ilike" else "ilike"
```
marielejeune: If we don't find a better solution for `ilike` case we can remove it from here, the domain modification is useless.

PR #700, opened 2023-01-27
[CHG] mozaik_mandate: allow to use wizard to close/reactivate several mandates together

marielejeune:
refs #65915

### DIFF
--- a/mozaik_mandate/tests/test_update_mandate_end_date_wizard.py
+++ b/mozaik_mandate/tests/test_update_mandate_end_date_wizard.py
@@ -102,7 +102,9 @@ class TestUpdateMandateEndDateWizard:
         wiz_id = wizard_object.with_context(context).create(
             {"mandate_deadline_date": next_month}
         )
-        self.assertEqual(wiz_id.message, "The selected mandate is already active!")
+        self.assertEqual(
+            wiz_id.message, "Some of the selected mandates are already active!"
+        )
 
 
 class TestUpdateStaMandateEndDateWizard(TestUpdateMandateEndDateWizard, SavepointCase):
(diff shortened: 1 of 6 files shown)


marielejeune:
@RobinetDenisAcsone do you know why this wizard was only dealing with a single mandate?

mozaik_mandate/wizards/update_int_mandate_end_date_wizard.xml
```diff
@@ -40,7 +42,7 @@
             name="groups_id"
             eval="[(6, 0, [ref('mozaik_abstract_model.res_groups_archiving_manager')])]"
         />
-        <field name="context">{'mode':'reactivate'}</field>
+        <field name="context">{'mode':'reactivate', 'active_test': False}</field>
```
marielejeune: If we do not pass this context key, `default_get` returns the ids of the mandates to reactivate but as they are archived, they are not taken into account in `mandate_ids` field of the wizard.
@RobinetDenisAcsone do you have a better solution?

RobinetDenisAcsone: Perhaps try something like `mandate_ids = fields.Many2many(comodel_name="ext.mandate", context={'active_test': False})`?
Not quite sure if this works.

RobinetDenisAcsone:
@marielejeune I don't know it was already like that in v8

mozaik_mandate/wizards/update_int_mandate_end_date_wizard.xml
marielejeune: Good idea, it works, thanks!

PR #714, opened 2023-03-14
[IMP] Run copier update

marielejeune:


.github/workflows/pre-commit.yml
```diff
@@ -11,7 +11,7 @@ on:
 
 jobs:
   pre-commit:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
marielejeune, line 14: @sbidoul running `copier update` changed this line. Should it stay?

.pre-commit-config.yaml
```diff
@@ -44,6 +44,11 @@ repos:
         entry: found forbidden files; remove them
         language: fail
         files: "\\.rej$"
+      - id: en-po-files
+        name: en.po files cannot exist
+        entry: found a en.po file
+        language: fail
+        files: '[a-zA-Z0-9_]*/i18n/en\.po$'
```
marielejeune, line 51: @sbidoul also not sure why we have to reject en.po files

sbidoul: @marielejeune because `en` is the language in the code, and when `en.po` files are present, bad things happen. See https://github.com/OCA/oca-addons-repo-template/issues/176.

.github/workflows/pre-commit.yml
sbidoul: Yes, better pin the version to avoid surprises when GitHub changes the OS of it's default runner image.

PR #721, opened 2023-04-07
[FIX] mozaik_sponsorship: sponsored membership but partner wants to pay more

marielejeune:
Fix special cases: 

* New member is sponsored, but he wants to pay his membership: the mozaik workflow created the membership line with the default product price (here 0 for a sponsored membership), and then modified the price (to copy the price of the MR) afterwards. But for sponsored memberships, `advance_workflow_as_paid` is  True on the product, hence the membership line was marked as paid and the workflow continued before changing the ML price. The result was a paid membership line but already marked as paid. 

* Existing member candidate now became sponsored. The MR sets his sponsor and the price is then 0. We overrode the `_validate_request_membership` method in `mozaik_sponsorship` but it was dealing with all active ML, even if the partner wasn't a member candidate before. The problem is that we recognized the sponsored product on basis of the price, and we set back the default price. But when we modified the default price by setting it in the MR, it was undone.
-> The `_validate_request_membership` overwritting must only be used for existing ML.

* When modifying the price of an existing membership line, trigger the workflow if the member candidate line price became 0 (because the ML is thus now paid)

Add a boolean `is_sponsored` on membership lines
This checkbox will be True on the active ML after validating the MR (hence not the member candidate line, but the following one).

### DIFF
--- a/mozaik_sponsorship/models/membership_request.py
+++ b/mozaik_sponsorship/models/membership_request.py
@@ -65,7 +65,7 @@ class MembershipRequest(models.Model):
             mr.partner_id.sponsor_id = mr.sponsor_id
         return res
 
-    def _validate_request_membership(self, partner):
+    def _validate_request_membership(self, partner):  # noqa: C901
         """
         When validating the MR, if the partner already has an active membership line
         in the same state as the MR result state, we do not create a new membership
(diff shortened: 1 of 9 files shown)


mozaik_sponsorship/models/membership_request.py
```diff
@@ -62,7 +62,15 @@ def validate_request(self):
         for mr in mr_can_be_sponsored.filtered(
             lambda mr: mr.partner_id and mr.sponsor_id
         ):
+            # Copy the sponsor on the partner
             mr.partner_id.sponsor_id = mr.sponsor_id
+            # Mark membership lines as sponsored membership lines, for instances
+            # in the MR
```
qgroulard: Explain why, not how.

The code is clear and does not need comments to explain what is done here.
In the other hand, it may need comments explaining why this is done.

marielejeune: This is just for reporting purposes...

qgroulard: What do you mean ?

marielejeune: We just added a boolean on membership lines and we are ticking it when validating the MR. 
I don't see what I could explain to answer to the "why?" question, as this new boolean was just added for reporting.

qgroulard: Ok sorry I thought you were justifying the need for the comments.

Then we don't need any comment at all.

PR #754, opened 2023-09-15
[ADD] mozaik_mass_mailing_sending_control: control before sending big mailings

marielejeune:
refs #70080

### DIFF
--- a/mozaik_mass_mailing_sending_control/wizards/mozaik_mass_mailing_sending_control.py
+++ b/mozaik_mass_mailing_sending_control/wizards/mozaik_mass_mailing_sending_control.py
@@ -1,0 +1,23 @@
+# Copyright 2023 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import _, fields, models
+from odoo.exceptions import ValidationError
+
+
+class MozaikMassMailingSendingControl(models.TransientModel):
+
+    _name = "mozaik.mass.mailing.sending.control"
+
+    mailing_id = fields.Many2one("mailing.mailing", required=True)
+    number_recipients = fields.Integer(readonly=True)
+    number_control = fields.Integer()
+
+    def doit(self):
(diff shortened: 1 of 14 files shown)


mozaik_mass_mailing_sending_control/wizards/mozaik_mass_mailing_sending_control.py
```diff
@@ -0,0 +1,23 @@
...
+    number_control = fields.Integer()
+
+    def doit(self):
+        self.ensure_one()
+        if self.number_recipients != self.number_control:
+            raise ValidationError(_("Please enter the correct number of recipients"))
+        if self.env.context.get("sending_operation") == "put_in_queue":
+            return self.mailing_id.with_context(control_done=True).action_put_in_queue()
+        if self.env.context.get("sending_operation") == "schedule":
+            return self.mailing_id.with_context(control_done=True).action_schedule()
```
qgroulard: I think calling again `action_put_in_queue()` and `action_schedule()` is a bad idea.
IMO the correct approach is to replace the buttons for queuing and scheduling by buttons that redirect to the check then call the previous actions.

The problem I see is if we have another module with a method like:
```
def action_put_in_queue(self):
    do_something()
    super().action_put_inqueue()
```

And this is called before your override for the check, then `do_something()` will be executed again when you call `self.mailing_id.with_context(control_done=True).action_put_in_queue()`.

qgroulard (changes requested):

mozaik_mass_mailing_sending_control/wizards/mozaik_mass_mailing_sending_control.py
marielejeune: You're right. But when replacing the buttons, we are going to loose all overrides that could have been added (attrs, confirm, ...). 

For example the domain of button `action_put_in_queue` is overridden in `mass_mailing_sms`

qgroulard (approved):

PR #758, opened 2023-09-22
[REF] New module mozaik_involvement_donation

marielejeune:
Extract the logic linked to donations in a new module. Add a configuration in `mozaik_membership_request` to specify which involvement types can be created from MR (by default donation cannot).

* Keep donations reconciliation in mozaik_account for now. It could be a nice idea to simplify this module too, but I think it won't be a pleasure.

NB: This PR only refactors existing features. New features asked in https://acsone.plan.io/issues/70158 will be done in a second PR.


qgroulard (approved):

marielejeune:
@qgroulard could you have a look at the two last commits please?

mozaik_membership_request/models/membership_request.py
```diff
@@ -276,6 +294,28 @@ def _get_status_values(self, request_type, date_from=False):
...
+    def _check_involvement_category(self):
+        """
+        Ensure that no forbidden involvement can be added on a MR.
+        Otherwise, the partner.involvement is created and then there is
+        a psycopg2 CheckViolation error on partner.involvement model (Odoo bug)
+        """
+        domain = self._get_involvement_category_domain()
+        allowed_ic_ids = set(
+            self.env["partner.involvement.category"]
+            .search(safe_eval.safe_eval(domain))
```
qgroulard: I think it would be better if `_get_involvement_category_domain()` returned a list and not a string.
Especially when playing with evals, it's better to keep it in one place.

Plus, you could use osv.expression to build the domain rather than slicing the string with things like `[1:-1]`.

PR #759, opened 2023-09-22
Donation via involvements : payment link and paying from website

marielejeune:
refs #70158

Depends on https://github.com/mozaik-association/mozaik/pull/760 

Pay attention that for not changing existing behavior, new modules 
`mozaik_involvement_donation` and `mozaik_account_donation` must be installed on all databases, hence added in `mozaik_all`.

Module `mozaik_involvement_donation_payment` adds new features (and related REST module) -> not in `mozaik_all`

**Also to review:** 2 last commits of https://github.com/mozaik-association/mozaik/pull/317, concerning REST services

**TODO**

- [x] Check that the payment date is written on the involvement only when the `payment.transaction` is completed (wire transfer case for e.g.)
- [x] Payment acquirer should be required on `partner.involvement.category` if `involvement_type='donation'`. We add the constraint on view only.
- [x] At each donation payment, an `account.move` is created, but the credit account is wrong. Allow to set this account (via config for e.g.)
- [x] Configure a "thanks" page after donation payment -> _Ok, it will be done manually on each payment acquirer._
- [x] Add `payment_link` as response of REST service creating a donation and extract everything linked to donations into a new rest module `mozaik_involvement_donation_rest`.
- [x] Add `copy=False` on `payment_date` on involvements


mozaik_involvement_donation/models/partner_involvement.py
```diff
@@ -31,3 +32,8 @@ def _compute_promise(self):
                 involvement.involvement_type == "donation"
                 and not involvement.payment_date
             )
+
+    @api.depends("promise")
+    def _compute_is_paid(self):
```
qgroulard, line 37: Is that correct ? All involvements are either paid or promise ?

mozaik_involvement_donation_payment/models/payment_transaction.py
```diff
@@ -38,4 +38,18 @@ def write(self, vals):
...
+            if len(self) == 1 and self.involvement_id:
+                # Update the account set on the related credit account move line.
+                # The account to take is configured by the user.
+                credit_line = (
+                    self.env["account.payment"]
+                    .browse(vals["payment_id"])
+                    .mapped("move_id.line_ids")
+                    .filtered(lambda t: t.credit > 0)
+                )
+                donation_account_id = self.env.user.company_id.donation_account_id
```
qgroulard: Please try to only name identifiers '..._id'. Record(set)s should not be named '..._id'.

mozaik_involvement_donation_payment/models/payment_transaction.py
```diff
@@ -38,4 +38,18 @@ def write(self, vals):
...
+                if donation_account_id:
+                    credit_line.account_id = int(donation_account_id)
```
qgroulard: Why do you need the parsing ?

marielejeune: Forgot to remove.

mozaik_involvement_donation/models/partner_involvement.py
marielejeune: No, you're right. Only donations can be marked as paid. Fixed in a new commit.

PR #760, opened 2023-09-27
[REF] Involvement donations

marielejeune:
**Only refactoring existing features**

- [x] Extract donations's logic present in `mozaik_account` into a new module `mozaik_account_donation`
- [x] Extract JS extensions of `payment` into a new module `mozaik_payment` as it will be common to membership and donation payments.
- [x] Extract `display_reference` into this common module `mozaik_payment`. This computed reference will be used on website payment pages for memberships and donations.
- [x] Extract logic about donations from `mozaik_virtual_partner_involvement` into a new module `mozaik_virtual_partner_involvement_donation`

**Add new logic not depending on donations**

- [x] Add a configuration on `payment.acquirer`. By default each payment acquirer can be used to pay memberships, but this option can be removed (it will be useful when adding acquirers that are specifically dedicated to donations)


mozaik_membership_payment/static/src/js/payment_form.js
```diff
@@ -7,230 +7,14 @@ odoo.define("mozaik_membership_payment.payment_form", function (require) {
     var _t = core._t;
 
     var paymentForm = require("payment.payment_form");
+    var origGetTransactionParams = paymentForm.prototype._get_transaction_params;
     paymentForm.include({
```
qgroulard, line 11: Could you please try to use `extend` and `super` here ?

`include` is for patch, `extend` is for inheritance.
https://www.odoo.com/documentation/14.0/fr/developer/reference/javascript/javascript_reference.html#inheritance

qgroulard (changes requested):

qgroulard (approved):

PR #761, opened 2023-10-20
[FIX] Event & Petitions: do not add involvement for tickbox questions not ticked

marielejeune:


mozaik_petition_membership_request_involvement/models/petition_registration.py
```diff
@@ -74,14 +74,15 @@ def _add_involvements_to_membership_request(self, request):
             ]
 
         # 3
-        for question in self.registration_answer_ids.question_id.filtered(
-            lambda s: s.involvement_category_id
```
qgroulard, line 78: Why was it that way ? Only to handle tickbox questions ?

It looks right to remove this part to me, but still I am wondering why it was added and if we'll miss something by removing it.

marielejeune: The idea was do stay generic in the case where other question types would be added. But I think, with this bug, that is not a wise idea.

PR #769, opened 2024-05-24
[FIX] mozaik_mandate: 'all details' action with wrong active_id

marielejeune:
As for other actions, there is a mix up with the ID and the active_id when having a long ariane path.
Converting the 'action' button into an 'object' button helps preventing this

refs #76024


marielejeune:
@qgroulard FYI. This will have to be merged into mozaik and the sources of mozaik_mandate must be updated on all mozaik projects.

PR #772, opened 2024-10-25
[ADD] mozaik_privacy_domain_mixin

marielejeune:
This was into a specific Cdh addon (cdh_ama_objects_domain) and the code was copied on events, petitions and survey.
* Extract it into a mixin
* Put it into Mozaik as it may serve (and it will very soon, for documents, next PR)
* `is_domain_set` is computed non stored, add a `search` method and add the `filter_allowed_for_partner` method
* Add unit tests.

### DIFF
--- a/mozaik_privacy_domain_mixin/__init__.py
+++ b/mozaik_privacy_domain_mixin/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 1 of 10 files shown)


marielejeune:
I need this to be merged but @qgroulard your review is still welcomed.

PR #780, opened 2024-12-12
[FIX] mozaik_partner_global_opt_out: include opt-out for mailing lists

marielejeune:
Fix the process uncanceling emails that were canceled because of a global opt-out.
The case where the mailing model is mailing.contact wasn't covered yet.

@qgroulard Mozaik being Mozaik ...
image


qgroulard (approved):
At least the patch is clean :laughing:

PR #792, opened 2025-03-11
[FIX] mass_mailing_distribution_list:

marielejeune:
Due to Odoo sources update, reply_to_mode computation changed It was set as 'thread' for distribution.list mailings, but it should be kept as 'email'.
See Odoo change in commit: da787012c1b6774d01f25c9dcc42d6161730ed16


marielejeune:
Ping @qgroulard :
- Fix mass mailings due to Odoo sources update
- Upgrade of the github templates due to a deprecation for actions/cache@v1

PR #796, opened 2025-06-16
[IMP] mozaik_event_barcode: computation of event domain was too slow

marielejeune:


mozaik_event_barcode/models/event_event.py
```diff
@@ -45,29 +45,11 @@ def _get_voting_partners(self):
...
-        """,
-            {
-                "vt_list": tuple(virtual_target_ids),
-            },
-        )
-        res_ids = [r[0] for r in self.env.cr.fetchall()]
-        if res_ids:
-            return (
-                self.env["res.partner"].search([("id", "in", res_ids)]).ids
-            )  # apply security
```
marielejeune, line 69: Not sure about this, but I guess that now that the partner ids are obtained with a map, the security is applied?
Moreover I don't really see why security is needed, this method must return the list of partners verifying the event's domain, to check if the partner we are registrating to the event is allowed to vote or not.

qgroulard (approved):

PR #798, opened 2025-08-01
[FIX] privacy_domain_mixin: performance

sbidoul:
By asking postgres to filter on the partner id, we use the primary key and avoid returning and filtering on many partners.

In future work we should look at avoiding
this search-inside-a-loop.


qgroulard:
FYI @marielejeune

qgroulard (approved):

PR #799, opened 2025-08-02
[FIX] mozaik_document: do not access content to get its size

sbidoul:
Reading the content from disk is a costly operation, and even more so using object storage.


sbidoul:
It turns out https://github.com/mozaik-association/mozaik/pull/798 was only a part of the solution.

qgroulard:
@marielejeune

PR #806, opened 2026-01-30
[14.0][ADD] Membership Request Origin

qgroulard:


mozaik_membership_request_origin/models/membership_request_origin.py
```diff
@@ -0,0 +1,20 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import _, fields, models
+from odoo.exceptions import UserError
+
+
+class MembershipRequestOrigin(models.Model):
+    _name = "membership.request.origin"
+    _description = "Membership Request Origin"
+    _inherit = ["mozaik.abstract.model"]
```
marielejeune: Do you need to inherit from this abstract model?
I don't like this abstract model, it lead to problems on another PR, I would like to use it as less as possible ...

mozaik_membership_request_origin/views/membership_request.xml
```diff
@@ -0,0 +1,17 @@
...
+<odoo>
+    <record model="ir.ui.view" id="membership_request_form_view">
+        <field name="model">membership.request</field>
+        <field
+            name="inherit_id"
+            ref="mozaik_membership_request.membership_request_form_view"
+        />
+        <field name="arch" type="xml">
+            <group name="request_info" position="inside">
+                <field name="origin_id" />
```
marielejeune: I would add a `readonly="1"` here.
It shouldn't be modifiable by a user.

mozaik_membership_request_origin/views/membership_request_origin.xml
```diff
@@ -0,0 +1,44 @@
...
+                </sheet>
+            </form>
+        </field>
+    </record>
+
+
+    <record model="ir.ui.view" id="membership_request_origin_tree_view">
+        <field name="model">membership.request.origin</field>
+        <field name="arch" type="xml">
+            <tree>
```
marielejeune: NIT: Please add `editable="bottom"`, it's more comfortable for the user.

mozaik_membership_request_origin/models/membership_request_origin.py
```diff
@@ -0,0 +1,20 @@
...
+    _unicity_keys = "name"
+
+    name = fields.Char(required=True, translate=True)
```
marielejeune: Isn't it a bit strange to add a unicity constraint on a translatable field? The unicity will only be checked in English, isn't it? Nothing prevents the user to put several times the same translation.

mozaik_survey_membership_request_origin/models/survey_user_input.py
```diff
@@ -0,0 +1,17 @@
...
+    _inherit = "survey.user_input"
+
+    @api.model
+    def create(self, vals):
+        res = super().create(vals)
+        survey_origin = self.env.ref(
+            "mozaik_survey_membership_request_origin.membership_request_origin_survey"
+        )
+        res.mapped("membership_request_id").write({"origin_id": survey_origin.id})
+        return res
```
marielejeune, line 17: NB: I'm wondering why the process of creating a MR from a survey user input doesn't follow the same schema as event registrations and petition signatures (using `_{create/update}_membership_request_from_registration`).
It would have been more coherent 😅

marielejeune (changes requested):
Minor comments, otherwise looks quite good!

mozaik_membership_request_origin/views/membership_request.xml
qgroulard: Why do we let the users create new origins if we don't allow to add them to membership requests ?

mozaik_membership_request_origin/models/membership_request_origin.py
qgroulard: No need I think, I just took inspiration from other mozaik modules.

I agree, I'll get rid of this and keep it simple.

mozaik_membership_request_origin/views/membership_request.xml
marielejeune: They will create a few new origins and set them when creating a MR from REST (examples: payment, address change, ...). 
But we can let the field modifiable for now and they'll notify us during their tests if we need to change it.

qgroulard: It makes sense, I have set it readonly.

marielejeune (approved):

PR #810, opened 2026-03-23
[14.0][ADD] mozaik_event_registration_counter

qgroulard:


mozaik_event_registration_counter/models/event_registration.py
```diff
@@ -0,0 +1,48 @@
...
+    )
+
+    def default_get(self, fields_list):
+        """
+        We want this logic to apply only from the view, hence the use of 'default_get()'.
+        """
+        vals = super().default_get(fields_list)
+        if "event_id" in vals:
+            event = self.env["event.event"].browse(vals["event_id"]).exists()
+            if event and event.date_begin < fields.Datetime.now():
```
marielejeune, line 23: Wondering if we shouldn't add a res.config.setting to be able to activate this option x minutes before the start of the event (we could imagine they open the doors x minutes before the event starts and they start encoding the live registrations).

mozaik_event_registration_counter/models/event_registration.py
```diff
@@ -0,0 +1,48 @@
...
+            for mail_scheduler in rec.event_id.event_mail_ids.filtered(
+                lambda scheduler: scheduler.interval_type == "after_sub"
+            ):
+                self.env["event.mail.registration"].create(
+                    {
+                        "scheduler_id": mail_scheduler.id,
+                        "registration_id": rec.id,
+                        "mail_sent": True,
+                    }
+                )
```
marielejeune: Prepare all create vals and call the create method only once?

mozaik_event_registration_counter/models/event_registration.py
marielejeune: not needed for now, it doors open a bit before the event, they can toggle the button manually. Let's keep it like that for now.

marielejeune (approved):
One minor improvement comment otherwise looks good!

PR #812, opened 2026-03-25
[14.0][ADD] mozaik_event_registration_partner_identifier

qgroulard:


test-requirements.txt
```diff
@@ -1,4 +1,3 @@
-odoo14-addon-base-view-inheritance-extension @ git+https://github.com/OCA/server-tools.git@14.0=setup/base_view_inheritance_extension
```
sbidoul: missing `#subdirectory=`

marielejeune (changes requested):
I would add this module as a dependency of `mozaik_all`, I don't see why some parties would refuse this improvement.

qgroulard:
> I would add this module as a dependency of `mozaik_all`, I don't see why some parties would refuse this improvement.

Well I thought `mozaik_all` was a mistake not to be used anymore.

PR #818, opened 2026-04-01
[ADD] mozaik_territory

baimont:
@marielejeune


mozaik_territory/views/territory.xml
```diff
@@ -0,0 +1,74 @@
...
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+
+    <record id="territory_view_tree" model="ir.ui.view">
+        <field name="name">territory.tree</field>
+        <field name="model">territory</field>
+        <field name="arch" type="xml">
+            <tree string="Territories">
+                <field name="name" />
+                <field name="active" invisible="1" />
```
marielejeune: Do you need the field in the tree view?

baimont: you're right, not necessary on tree view

PR #821, opened 2026-04-09
[FIX] mozaik_person: name_search on res.partner: allow identifiers starting with 0

marielejeune:
refs #94547


qgroulard:
So I guess the problem was only due to leading 0 ?

qgroulard:
> So I guess the problem was only due to leading 0 ?

Oups, didn't read the title of the PR and commit :see_no_evil:

